PR #50, opened 2023-08-27
Added Functional Test Codes for Python Django and Fixed Regex Code.

ksg97031:
I have updated the following tasks:

- Added functional test codes for Python Django.
- Fixed a bug in the Django Analyzer.
- Updated the .gitignore file.


hahwul (approved):
Cool 😉

PR #170, opened 2023-11-20
Add cookie type of param, Improve logger and baker

hahwul:


ksg97031 (approved):
I've included the commit that addresses the cookie parameter handling in the Python web analyzer!
Thanks a lot! 👍👍

Issue #184, opened 2023-12-08
Implement Common URL Handling Logic


hahwul:
Currently, we are combining the Endpoint and URL in the Analyzer. However, this approach allows for disparate handling in each Analyzer, so I plan to centralize URL handling as common logic.

PR #188, opened 2023-12-09
Implement Common URL Handling Logic (from #184)

hahwul:


hahwul:
Issue: #184

src/analyzer/analyzers/analyzer_spring.cr
ksg97031, line 1: What if we replace url with an empty string ("") instead of using @url?

https://github.com/noir-cr/noir/blob/ffd42a0349f180cacf821ec768730f125d00c862/src/analyzer/analyzers/analyzer_spring.cr#L14

spec/unit_test/models/noir_spec.cr
ksg97031, line 1: Could we handle cases where the path doesn't start with a '/'? 
For example, `tmp_endpoint2 = Endpoint.new("abcd", "GET")`

ksg97031 (approved):
I've appended a commit to refactor the combine_url_endpoints method!

Issue #192, opened 2023-12-10
Noir for VSCode Extension 


ksg97031:

ksg97031:
vscode command for file navigation (https://code.visualstudio.com/api/references/commands)
```typescript
vscode.commands.executeCommand(
  "editor.action.goToLocations",
  vscode.Uri.file(filepath),
  new vscode.Position(linenumber, 0),
  [],
  "goto"
);
```

ksg97031:
Hello @hahwul !

I've added a new option to include file paths in the result output. I did this by adding a details field (a struct type) to the Endpoint to store the path and line number information. You can check out the changes in this commit:

https://github.com/noir-cr/noir/tree/include-path-option https://github.com/noir-cr/noir/commit/745652b0a1c3ebd6dd00276310cf96871a99f204

I'd appreciate your feedback on this approach. Do you think it's suitable, or are there better ways to achieve this functionality?

Here are examples of how the output looks with the new option now:
- with `--include-file-path` option
![image](https://github.com/noir-cr/noir/assets/6715194/3817a1ae-be19-4769-b979-db67503d94a4)

- with `-json` option
![image](https://github.com/noir-cr/noir/assets/6715194/561bdd21-1145-4e02-ae1a-849c245cac33)

Please give me feedback whenever you have time. No rush.

hahwul:
Hiya @ksg97031 
Wow! It''s awesome.

Regarding feedback, how about allowing the provision of multiple file paths instead of just one? Sometimes, an application's endpoint can be declared from multiple files, so it would be great to consider accommodating such cases by providing an array format for file paths.  I think it could contribute to greater flexibility and adaptability, in my opinion. What are your thoughts on this idea?

And I think the name 'Details' could potentially hold more than just 'File' and 'Line' information. So, I've made some improvements below, and I'm curious to hear your thoughts on this.

e.g
```crystal
struct Details
  include JSON::Serializable
  include YAML::Serializable
  @paths : Array(Path) = [] of Path
  # + New details types to be added in the future..

  def add_path(path : Path)
    @paths << path
  end
end

struct Path
  include JSON::Serializable
  include YAML::Serializable
  property path, line

  def initialize(@path : String, @line : Int32 | Nil)
  end
end
```

```json
[
  {
    "headers": [],
    "method": "POST",
    "params": [],
    "protocol": "http",
    "url": "https://testapp.internal.domains/comments",
    "details": {
       "paths": [
[...]

Issue #155, opened 2023-11-12
Improve Performance


hahwul:

hahwul:
As expected, using lightweight threads for each file in large quantities is much faster, likely due to reduced waiting time for lightweight threads caused by File I/O. However, there is a potential risk of simultaneous opening of numerous files. It's a matter of weighing the advantages and disadvantages to determine the optimal approach.

FIber per each file: 16s
Workerpool pattern: 38s

```crystal
channel = Channel(String).new
  spawn do 
    Dir.glob("#{base_path}/**/*") do |file|
      channel.send(file)
    end
  end

  500.times do 
    spawn do 
      loop do 
        begin
          file = channel.receive
          next if File.directory?(file)
          content = File.read(file, encoding: "utf-8", invalid: :skip)
  
          detector_list.each do |detector|
            if detector.detect(file, content)
              techs << detector.name
            end
          end
        rescue e : File::NotFoundError
          logger.debug "File not found: #{file}"
        end
      end
    end
  end

  Fiber.yield
  techs.uniq
```

PR #213, opened 2023-12-29
Add --include-path flag (#155)

ksg97031:
Thank you for the feedback provided at https://github.com/noir-cr/noir/issues/192#issuecomment-1868391647

I have incorporated your suggestions and made some adjustments:

1. Renamed the `Path` struct to `PathInfo`.
The [Path](https://crystal-lang.org/api/1.10.1/Path.html) struct had already been declared.

2. Changed the `--include-file-path` flag to `--include-path`.
I considered using --with-path, but it resembled the --with-headers flag and might cause confusion.

3. Replaced the use of the `paths` variable with `code_paths` in the code.
Using `paths` led to code like `paths[0].path`, which seemed a bit awkward. Therefore, I opted for `code_paths` as the variable name instead.

In the process of refactoring, several analyzer files now include the `Details` object for storing path and line number information. However, in a few cases where the line number couldn't be detected, only path information is stored.

I welcome any additional feedback!

### DIFF
--- a/spec/unit_test/models/endpoint_spec.cr
+++ b/spec/unit_test/models/endpoint_spec.cr
@@ -24,4 +24,41 @@ describe "Initialize 3 arguments" do
   it "detect_params" do
     endpoint.params.should eq([Param.new("a", "b", "query")])
   end
+
+  path = "path/a/b/c"
+  line = 123
+  path_info = PathInfo.new(path, line)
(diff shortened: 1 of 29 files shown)


hahwul (approved):
The code is neat and straightforward, making it a pleasure to read. No difficulties whatsoever. Approving this excellent work :D

Issue #232, opened 2024-01-23
Plans Regarding Spring (Java & Kotlin) Parameter Support


sm1ee:
Hello Noir Team,
This is a wonderful tool and I am looking forward to how far it will progress.
Since a lot of web frameworks now these days use Spring Framework, I was wondering when the parameter parsing for Java & Kotlin Spring Framework will be supported.
If you have a rough roadmap or outline, I would really appreciate it.

hahwul:
@ksg97031 👀

Dyrandy:
I agree.

hahwul:
@sm1ee 
Hello again 👋🏼
We've just released [v0.13.0](https://github.com/noir-cr/noir/releases/tag/v0.13.0), which includes enhancements to the analysis of Java Spring. We're planning improvements for Kotlin Spring in future versions. It appears advantageous to handle these issues separately, so I'll close the current one and create a new issue specifically for Kotlin Spring :D

Thank you so much for the contribution!!

> @ksg97031 I was impressed by your awesome code 🤩

PR #233, opened 2024-01-23
👔 Add identification logic of parameters in Spring

hahwul:
Hi @ksg97031 👋🏼

I've added some code to the SpringAnalyzer for #232. This code identifies well-known parameter processing logic and adds it as a parameter to the last added Endpoint during file content analysis. I've written it lightly for now, but please review and let me know if this method seems good!

If this direction is correct, I think it would be good to continue expanding the pattern together in the future. However, I'm slightly worried because developers have such a wide variety of ways to handle parameters.


ksg97031 (approved):

ksg97031:
@hahwul 

Hi~ I've implemented a simple lexer and parser for the Java language.
Now we can gather parameters for Spring MVC annotations and endpoints more accurately. 
Since it took longer than expected, I decided to implement it first for Java then Kotlin later. 
Let me know if you encounter any suspicious issues while using it!

hahwul:
@ksg97031 
Cool! Thank you for the great work! 
I'll make sure to try to include the changes you've applied in the next release (maybe 0.13.0). 

Have a wonderful refresh :D

Issue #330, opened 2024-06-19
Duplicate Entries in Added and Removed Lists in Diff Mode


ksg97031:
**Describe the bug**
When running the `OutputBuilderDiff` class to compare two sets of endpoints, the `added` and `removed` lists contain duplicate entries. This issue likely occurs because the `Endpoint` class does not implement the `==` and `hash` methods, leading to improper comparison and hashing of endpoint objects.

**To Reproduce**
1. Run the `OutputBuilderDiff` class with two sets of endpoint data.
2. Observe the `added` and `removed` lists in the output.
3. Notice that many endpoints are incorrectly listed as both added and removed, despite having the same URL, method, and parameters.

**Expected behavior**
The `added` list should contain only the endpoints that are present in the new set but not in the old set. The `removed` list should contain only the endpoints that are present in the old set but not in the new set. Endpoints that are unchanged should not appear in either list.

**Screenshots**
<img width="377" alt="image" src="https://github.com/noir-cr/noir/assets/6715194/4e6514dc-c61f-4d65-8a54-9a20d0042537">

**Versions**
- OS: macOS
- Version: v0.16.0

**Additional context**
The bug is likely due to the `Endpoint` class not implementing the `==` and `hash` methods. This may lead to endpoints with the same properties being treated as different objects. Implementing the `==` method to compare the URL, method, and parameters of the endpoints, and the `hash` method to generate a hash based on these properties, should ensure proper functionality in hash-based data structures.

hahwul:
@ksg97031 
I'll take care of it in the next version! Thank you for the report :)

PR #335, opened 2024-06-26
🧪 Implement endpoint comparison logic in diff analysis for #330

hahwul:
@ksg97031 
I made some changes based on what I understood. Please take a look and let me know if anything seems off or if the direction isn't quite right. Your feedback would be really helpful!


ksg97031 (approved):
LGTM! DiffMode is working correctly.

PR #394, opened 2024-09-14
Add path param to endpoint

ksg97031:
Hello @hahwul,

When you have time, I would appreciate it if you could review this.
Have a great Chuseok!!

Resolve: #389

### **Output**
Below is the output after the changes have been applied:
image

### **When Using the `--set-pvalue=test` Flag**
Here is the output when the `--set-pvalue=test` flag is set:
image

### Additional Suggestion:
Currently, the `--set-pvalue` flag is applied to the values of all parameters.
IMAO, It might be better to enhance this option to allow specifying the value for a particular parameter by its key.

### DIFF
--- a/spec/functional_test/func_spec.cr
+++ b/spec/functional_test/func_spec.cr
@@ -95,17 +95,17 @@ class FunctionalTester
             if endpoint.params.size > 0
               describe "check - params" do
                 endpoint.params.each do |param|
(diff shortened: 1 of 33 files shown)


hahwul (approved):

hahwul:
@ksg97031 
Wow, great work!!

hahwul:
@ksg97031 그리고 즐거운 추석 보내세요 :D

Issue #395, opened 2024-09-15
Improvement: Extend --set-pvalue Functionality for Cookies, Headers, Queries and ETC


hahwul:
> From this discussion https://github.com/owasp-noir/noir/pull/394#issue-2526429180

@ksg97031 
How about handling `--set-pvalue` like this?

--set-pvalue: Set values for all parameters
--set-pvalue-cookie: Set values specifically for cookie parameters
--set-pvalue-header: Set values specifically for header parameters
--set-pvalue-query: Set values specifically for query parameters
etc...

It's just an idea, but is there a better way to do this?

ksg97031:
Yes, I also think separating the options by parameter types is a good approach.
Additionally, how about supporting a key/value pair format for `set-pvalue*` parameters?

For example:

`--set-pvalue page=1 --set-pvalue test`
Structure: `/<page>?param={param}`
Result: `/1?param=test`

`--set-pvalue year=2024 --set-pvalue month=09`
Structure: `/<year>/<month>/calendar.html`
Result: `/2024/09/calendar.html`

`--set-pvalue-path id=395 --set-pvalue-query id=ksg97031`
Structure: `/posts/<id>?id={id}`
Result: `/posts/395?id=ksg97031`

Would it be feasible to implement it this way?

hahwul:
@ksg97031 
Sure, that's a great idea!! So let's roll up our sleeves and get it done🔥

PR #399, opened 2024-09-18
Refactor code to use YAML::Any for all options

hahwul:
Previously, we used complex option handling to maintain `Hash(String, String)`. Now, we use YAML-based Options(`Hash(String, YAML::Any)`), which allows for easy handling of various types and more.

For #398 
Refer: #395

스크린샷 2024-09-18 오후 6 57 32

### DIFF
--- a/docs/_advanced/configuration.md
+++ b/docs/_advanced/configuration.md
@@ -31,7 +31,7 @@ layout: page
 base: ""
 
 # Whether to use color in the output
-color: "yes"
+color: true
 
 # The configuration file to use
 config_file: ""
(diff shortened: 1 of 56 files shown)


hahwul:
Hi @ksg97031,

FYI:

The usage of Options has slightly changed. Now, instead of always being a `String`, it uses [`YAML::Any`](https://crystal-lang.org/api/1.13.2/YAML/Any.html). This means type casting is required when accessing these options.

For example:

```crystal
# Old way:
base_path = @options["base"]
puts @options["debug"]
# => no

# New way:
base_path = @options["base"].to_s
puts @options["debug"]
# => false
```

-----


*There are no more weird codes like this 😆*

PR #519, opened 2025-01-25
Enhancing Accuracy of LLM Integration

ksg97031:
@hahwul
I found the detailed format option!
Could you please review this PR?

Also, I have an idea:
How about changing the file filtering logic from using LLM to a standard search that looks for matching keywords (such as "endpoint," "url," "router," "controller," etc.) in the file?

저도 나가봐야해서 천천히 봐주십쇼!

### DIFF
--- a/src/llm/ollama/ollama.cr
+++ b/src/llm/ollama/ollama.cr
@@ -8,30 +8,13 @@ module LLM
       @model = model
     end
 
-    def request(prompt : String)
+    def request(prompt : String, format : String = "json")
       body = {
         :model  => @model,
         :prompt => prompt,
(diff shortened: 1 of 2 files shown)


hahwul:
@ksg97031 
Oh, that's a really good idea! I think that's a much more efficient way.

*연휴 잘 보내세요 😊*

hahwul (approved):

src/llm/ollama/ollama.cr
hahwul, line 1: ## Memo
https://github.com/ollama/ollama/blob/main/docs/api.md#parameters

PR #534, opened 2025-02-15
Add --help-all flag

hahwul:


hahwul:
Why does the test fail?

PR #541, opened 2025-02-19
feat: Add General AI Analyzer(OpenAI Compatibility API)

hahwul:


hahwul:
Hi @ksg97031 
Here’s a draft of the AI integration flags I’ve put together. I’ve added --openai, --openai-model, and --openai-api-key. These flags work not just with OpenAI but also with services like vLLM and LM Studio that support the OpenAI API format. I was wondering if these flag names are the right choice. I considered using something more generic like --inference (e.g., --inference-url), but I leaned toward --openai for consistency with the existing --ollama flags.

```
  AI Integration:
    --ollama http://localhost:11434  Set the URL for the Ollama server
    --ollama-model MODEL             Specify the name of the model to use with the Ollama server
    --openai http://api.openai.com   OpenAI-compatible API server URL
    --openai-model MODEL             Model name for OpenAI-compatible API
    --openai-api-key KEY             API key for OpenAI-compatible API
```

I think --openai makes sense since it clearly signals compatibility with OpenAI-style APIs. That said, a more generic name like --inference might be worth considering for broader flexibility. What do you think? Do you have any better ideas or suggestions for improvement?

Thanks!

hahwul:
@ksg97031 
The xAI API is also identical to the OpenAI spec. Since they basically use the same format, how about unifying the options and branching them through the provider? It seems like we could keep the ollama flag for now and deprecate it later.

`--ai-server`
`--ai-model`
`--ai-key`
`--ai-provider`

hahwul:
## Flags
```
  AI Integration:
    --ai-server PREFIX|URL           Set the URL for an AI(LLM) API server
                                       * Prefix
                                       * openai: http://api.openai.com
                                       * ollama: http://localhost:11434
                                       + e.g., --ai-server=openai , --ai-server=http://localhost:9100
    --ai-model MODEL                 Specify the model name for the OpenAI-compatible API
    --ai-key KEY                     Set the API key for the OpenAI-compatible API
    --ai-platform PLATFORM           Specify the AI platform (default: general)
    --ollama http://localhost:11434  Set the URL for the Ollama server
    --ollama-model MODEL             Specify the name of the model to use with the Ollama server
```

## Test
```bash
./bin/noir -b ./spec/functional_test/fixtures/hahwul \
                --ai-server ollama \
                --ai-model llama3
```

ksg97031:
> @ksg97031 The xAI API is also identical to the OpenAI spec. Since they basically use the same format, how about unifying the options and branching them through the provider? It seems like we could keep the ollama flag for now and deprecate it later.
> 
> `--ai-server` `--ai-model` `--ai-key` `--ai-provider`

Sorry for the late reply.
I agree with using a unified format and removing the ollama flag in the future.
Looks good to me!

src/analyzer/analyzers/llm_analyzers/general.cr
```diff
@@ -0,0 +1,106 @@
...
+          params = ep["params"].as_a.map do |param|
+            p = param.as_h
+            name = p["name"].as_s
+            param_type = if p.has_key?("param_type")
+                           p["param_type"].as_s
+                         elsif p.has_key?("type")
+                           p["type"].as_s
+                         else
+                           ""
+                         end
```
ksg97031, line 86: 이런 문법도 가능하군요!

ksg97031 (approved):
I've reviewed the code changes.
Personally, I think I need to try out the API integration a bit more.
Thanks for your hard work!!

hahwul:
## Request Body
### x.ai
```json
{
  "deferred": null,
  "frequency_penalty": null,
  "logit_bias": null,
  "logprobs": null,
  "max_tokens": null,
  "messages": [],
  "model": "",
  "n": null,
  "presence_penalty": null,
  "response_format": [
    null,
    [
      {
        "type": ""
      },
      {
        "type": ""
      },
      {
        "json_schema": null,
        "type": ""
      }
    ]
  ],
  "seed": null,
  "stop": null,
  "stream": null,
  "stream_options": [
    null,
    {
      "include_usage": false
    }
  ],
  "temperature": null,
  "tool_choice": [
    null,
    [
      "",
      {
        "function": [
          null,
          {
            "name": ""
          }
        ],
        "type": ""
      }
    ]
  ],
  "tools": null,
  "top_logprobs": null,
  "top_p": null,
  "user": null
}
```

PR #600, opened 2025-05-22
Fixed bug in max_tokens

hahwul:


hahwul:
Github action's bug...

-> https://github.com/owasp-noir/noir/actions/runs/15179390995

PR #627, opened 2025-07-07
fix: add url_paths check https://github.com/owasp-noir/noir/issues/626

EvtDanya:
I think we can fix this issue with this commit https://github.com/owasp-noir/noir/issues/626


EvtDanya:
image

test

ksg97031:
@EvtDanya
LGTM. Thank you for contributing to the Spring analyzer!
We'll fix the other issues as soon as possible. Thanks again!

PR #1078, opened 2026-03-09
detector(php): tighten Symfony heuristics and add detector specs

fabiencheret:
Hi team! This PR aims to better restrict Symfony detector heuristics

## Why
Previous checks could detect Symfony from weak signals (or string literals), which increased false positives.

## Changes
- Updated `Detector::Php::Symfony` rules:
  - `config/bundles.php` now requires `Symfony\` in content.
  - `config/services.yaml` now requires `App\` in content.
  - PHP detection now matches full real import statements: `use Symfony\...;`.
  - Supports one-line import style like `<?php use Symfony\...;`.
  - Removed annotation/attribute fallback (`@Route`, `#[Route]`) to avoid non-Symfony matches.

### DIFF
--- a/spec/unit_test/detector/php/symfony_detector_spec.cr
+++ b/spec/unit_test/detector/php/symfony_detector_spec.cr
@@ -28,10 +28,10 @@ describe "Detect Symfony" do
   end
 
   it "detects config/bundles.php with Symfony\\" do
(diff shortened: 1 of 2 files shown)


hahwul (approved):
LGTM! Great work!

PR #1110, opened 2026-03-17
Feat: add NOIR_PARSER_MAX_DEPTH to limit parser import depth

4ICH4:
## Summary

Adds an optional environment variable **`NOIR_PARSER_MAX_DEPTH`** so that users can cap how deep parsers follow imports. This keeps analysis time predictable on large codebases and in CI (e.g. Python/Flask, Java/Kotlin Spring), where miniparsers recursively resolve imports.

## Problem

On large repos, the Python miniparser (and similar logic in Java/Kotlin analyzers) can follow long import chains and parse a huge number of files. In practice this can mean:

- **Analysis runs for over an hour** or appears to hang.
- One entry file can pull in hundreds of modules (and their imports), so runtime grows quickly with repo size and import depth.

Without a way to cap how deep parsers follow imports, users on large codebases have no way to get a bounded, predictable run time.

## Solution

- **`NOIR_PARSER_MAX_DEPTH`** (integer):
  - **`0`** = only the entry file is parsed (no import following).
  - **`1`** = entry file + direct imports (one level).
  - **`2`** = entry + direct imports + their imports (two levels).
  - **Unset or negative** = no limit (previous behaviour).

Parsers stop following new imports once the current depth reaches the limit. Entry files are always at depth 0.

## Changes

| Area | Change |
|------|--------|
| **`src/utils/parser_limit.cr`** | New module that reads `NOIR_PARSER_MAX_DEPTH` and exposes `allow_depth?(depth)`. |
| **Python miniparser** | Tracks depth per parser instance; skips following imports when `depth >= max_depth`; root parser is created with depth 0 from Flask. |
| **Flask analyzer** | Creates the root Python parser with `depth: 0`. |
| **Java Spring & Kotlin Spring** | Before creating a parser for an imported or same-package file, call `ParserLimit.allow_depth?(0)` so depth-1 parsing respects the limit. |
| **Options / help** | `noir --help-all` documents `NOIR_PARSER_MAX_DEPTH` in the environment variables section. |

## Usage

```
# Entry files only (fastest; may miss endpoints that depend on imports)
NOIR_PARSER_MAX_DEPTH=0 noir -b /path/to/app

# Entry + direct imports (good default for CI)
NOIR_PARSER_MAX_DEPTH=1 noir -b /path/to/app

# No limit (default if unset)
noir -b /path/to/app

```

### DIFF
--- a/spec/unit_test/utils/parser_limit_spec.cr
+++ b/spec/unit_test/utils/parser_limit_spec.cr
@@ -1,0 +1,80 @@
+require "../../spec_helper"
+require "../../../src/utils/parser_limit"
+
+describe "ParserLimit" do
+  # Reset state before each test so ENV changes take effect.
+  before_each { ParserLimit.reset }
+  after_all { ENV.delete("NOIR_PARSER_MAX_DEPTH"); ParserLimit.reset }
+
+  describe ".max_depth" do
+    it "returns nil when NOIR_PARSER_MAX_DEPTH is not set" do
+      ENV.delete("NOIR_PARSER_MAX_DEPTH")
+      ParserLimit.max_depth.should be_nil
+    end
+
+    it "returns 0 when set to 0" do
+      ENV["NOIR_PARSER_MAX_DEPTH"] = "0"
+      ParserLimit.max_depth.should eq(0)
+    end
+
+    it "returns the integer value when set to a positive number" do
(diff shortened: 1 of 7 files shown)


hahwul (commented):
Nice feature! the Python depth tracking is solid.

hahwul (approved):

PR #1119, opened 2026-03-19
Fix Chi parameter extraction in mounted router functions

### LINKED ISSUE #1108: Chi: extract parameters from mounted router function inline handlers
## Summary

Chi's `analyze_router_function` does not extract parameters from inline handlers inside mounted router functions.

## Background

When Chi resolves a `r.Mount("/api", apiRouter())` call, it uses `analyze_router_function` to find endpoints inside `apiRouter()`. However, `process_route_line` only extracts the route path and HTTP method — it does not track inline handler context or extract parameters like `FormValue`, `Query().Get`, `Header.Get`, etc.

## Example

```go
func apiRouter() http.Handler {
    r := chi.NewRouter()
    r.Post("/users", func(w http.ResponseWriter, r *http.Request) {
        name := r.FormValue("name")  // ← not extracted
        _ = name
    })
    return r
}
```

Expected: `POST /api/users` with param `name (form)`
Actual: `POST /api/users` with no params

## Scope

- Extend `analyze_router_function` to track inline handler state (similar to the main `analyze` loop)
- Extract parameters within inline handlers using existing `get_param` patterns

## Related

- Discovered during #1106 (cross-file route analysis)
- Pre-existing limitation, not a regression

fabiencheret:
## Summary

- Extract inline parameter detection logic into a reusable `extract_params` method on the Chi analyzer
- Call `extract_params` from `analyze_router_function` so that parameters (`FormValue`, `Query().Get`, `Header.Get`, etc.) are detected inside mounted router inline handlers
- Track `last_endpoint` in `analyze_router_function` to associate extracted params with the correct endpoint
- Update cross-file test fixtures and expectations to cover query, form, and header parameters

Closes #1108

### DIFF
--- a/spec/functional_test/fixtures/go/chi_crossfile/api_routes.go
+++ b/spec/functional_test/fixtures/go/chi_crossfile/api_routes.go
@@ -9,7 +9,8 @@ import (
 func apiRouter() http.Handler {
 	r := chi.NewRouter()
 	r.Get("/users", func(w http.ResponseWriter, r *http.Request) {
-		w.Write([]byte("users list"))
+		page := r.URL.Query().Get("page")
+		_ = page
 	})
 	r.Post("/users", func(w http.ResponseWriter, r *http.Request) {
 		name := r.FormValue("name")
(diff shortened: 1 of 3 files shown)


hahwul (approved):
@fabiencheret 
LGTM! Clean refactoring, the shared extract_params method is a nice improvement over the duplicated inline logic.

> One minor note: single-line inline handlers (e.g. `r.Get("/x", func(...) { r.FormValue("y") })`) won't get params extracted since handler_brace_count hits zero immediately. Pre-existing behavior though, not a concern for this PR.

PR #1130, opened 2026-03-26
Fix/cache error logging

### LINKED ISSUE #1124: Add error logging to bare rescue blocks in llm/cache.cr
## Description

`src/llm/cache.cr` has multiple bare `rescue` blocks (around lines 76, 85, 94, 107, 126, 142) that silently return `nil` or `false` when cache operations fail.

For example:
```crystal
begin
  # cache read/write operation
rescue
  nil
end
```

LLM cache failures being silently ignored can lead to:
- Repeated API calls when cache reads fail (cost and latency impact)
- Lost cache entries when writes fail (no indication to the user)
- Difficult debugging when LLM features behave unexpectedly

### Suggested fix

Capture and log exceptions at debug level:

```crystal
rescue e
  @logger.debug "Cache operation failed: #{e.message}"
  nil
end
```

oussama134:
Right now, the LLM cache module has bare rescue blocks that just swallow exceptions and return nil or false without telling anyone what went wrong. That makes it hard to debug when cache reads or writes fail, and can lead to unnecessary API calls or lost cache entries.

What I changed
Replaced every bare rescue with rescue e so we can actually see the error.
Added Log.debug messages to give us visibility into what failed.
Kept all the existing return values (nil, false, etc.) – so the behavior is exactly the same, just with logging.

Why it’s helpful
No functional changes, but now if something goes wrong with the cache, we’ll see a clear debug log instead of just silence. That’ll make it way easier to figure out issues when they happen.

### DIFF
--- a/src/llm/cache.cr
+++ b/src/llm/cache.cr
@@ -73,7 +73,8 @@ module LLM
       path = path_for(key)
       return unless File.exists?(path)
       File.read(path)
-    rescue
+    rescue e
+      Log.debug { "Cache fetch failed for #{key}: #{e.message}" }
       nil
     end
 
@@ -82,7 +83,8 @@ module LLM
       ensure_dir
       File.write(path_for(key), content)
       true
-    rescue
+    rescue e
+      Log.debug { "Cache store failed for #{key}: #{e.message}" }
       false
     end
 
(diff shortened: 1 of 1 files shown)


hahwul (approved):
LGTM!

Nice improvement! logging LLM cache errors instead of silently swallowing them. Clean and non-breaking. Approved. Thanks!!

PR #1142, opened 2026-03-31
fix: preserve accurate line numbers for JS endpoints

### LINKED ISSUE #1141: JS analyzers always report line 1 for detected endpoints
## Description

When scanning JavaScript services (Express/Koa/Fastify/Restify), Noir reports endpoints with line 1 even when routes are defined later in the file. This makes results harder to triage and reduces trust in endpoint metadata.

## Steps to Reproduce

1. Create an Express app with routes defined past line 1, e.g.:

```js
const express = require('express');
const app = express();

// some middleware setup...

app.get('/users', (req, res) => {
  res.json([]);
});

app.post('/users', (req, res) => {
  res.status(201).send();
});
```

2. Run noir against the project
3. Observe that all detected endpoints report `line: 1` regardless of their actual position in the file

## Expected Behavior

Each endpoint should report the line number where the route is actually defined.

## Actual Behavior

All endpoints report `line: 1`.

4ICH4:
## Summary

When scanning JavaScript services, Noir was unconditionally overwriting endpoint details set by `JSRouteExtractor` with hardcoded approximate values (`line 1`), causing all detected endpoints to report the wrong source location regardless of where the route was actually defined in the file.

## Changes

- **`src/miniparsers/js_route_extractor.cr`**: Compute the accurate line number from `pattern.start_pos` by counting newlines in the content before that offset, and attach it to the endpoint when constructing it.
- **`src/analyzer/analyzers/javascript/express.cr`**: Only set `endpoint.details` if `code_paths` is empty, so the accurate line number set by the extractor is preserved instead of being overwritten.

## Testing

Verified that endpoints detected in Express apps now report the correct line number matching the route definition in the source file.

Closes #1141

### DIFF
--- a/src/analyzer/analyzers/javascript/express.cr
+++ b/src/analyzer/analyzers/javascript/express.cr
@@ -36,9 +36,10 @@ module Analyzer::Javascript
               content = File.read(path, encoding: "utf-8", invalid: :skip)
               parser_endpoints = Noir::JSRouteExtractor.extract_routes(path, content, @is_debug)
               parser_endpoints.each do |endpoint|
-                # Add file location details
-                details = Details.new(PathInfo.new(path, 1)) # Line number is approximate
-                endpoint.details = details
+                # Use the line number already set by the extractor; fall back to path-only if missing
+                if endpoint.details.code_paths.empty?
+                  endpoint.details = Details.new(PathInfo.new(path))
+                end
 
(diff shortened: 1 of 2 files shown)


hahwul (approved):
LGTM. Nice fix for the line number issue.
Thanks for addressing the JS endpoint accuracy. I'll merge it

PR #1153, opened 2026-04-03
Remove empty `Details#initialize` method in endpoint.cr

### LINKED ISSUE #1151: Remove empty `Details#initialize` method in endpoint.cr
## Description

`src/models/endpoint.cr` (lines 130–131) defines an empty `initialize` method for the `Details` struct:

```crystal
struct Details
  include JSON::Serializable
  include YAML::Serializable
  property code_paths : Array(PathInfo) = [] of PathInfo
  property status_code : Int32?
  property technology : String?

  # + New details types to be added in the future..

  def initialize
  end
  ...
```

Since all properties already have default values (empty array and `nil`), Crystal will auto-generate a no-arg initializer. The explicit empty `def initialize; end` is redundant.

Additionally, the comment on line 128 has a minor formatting issue — trailing double periods (`..`) and a `+` prefix.

### Suggested fix

```diff
- # + New details types to be added in the future..
-
- def initialize
- end
```

If the comment is still relevant, clean it up:
```crystal
# New details types can be added in the future.
```

tmchow:
Crystal structs auto-generate a default initializer when all properties have default values. The explicit empty `initialize` in `Details` is redundant since `code_paths` defaults to an empty array and `status_code`/`technology` are nilable. The parameterized `initialize(code_path)` overload is kept.

This contribution was developed with AI assistance (Claude Code).

Fixes #1151

### DIFF
--- a/src/models/endpoint.cr
+++ b/src/models/endpoint.cr
@@ -127,9 +127,6 @@ struct Details
 
   # + New details types to be added in the future..
 
-  def initialize
-  end
-
   def initialize(code_path : PathInfo)
     @code_paths << code_path
   end


tmchow:
Closing this - the empty `initialize` is actually required. Crystal doesn't auto-generate a no-arg initializer when a parameterized `initialize` overload exists, so `Details.new` (with no arguments) fails without it. The CI failure confirms: `no overload matches 'Details.new'`. Sorry for the noise.

PR #1230, opened 2026-04-17
chore: switch to official crystallang/crystal image

hahwul:
## Summary
- Crystal now provides [official Linux ARM64 builds](https://crystal-lang.org/2026/04/07/official-linux-arm64-builds/), so we can drop the 84codes images that were previously needed for multi-arch coverage.
- Updated `Dockerfile`, the release-binaries workflow, and `AGENTS.md` to use `crystallang/crystal:1.19.1`.

## Test plan
- [x] `docker build .` succeeds on linux/arm64 with static linking
- [x] Built image runs: `noir --version` → `0.29.1`
- [x] Built image runs: `noir -b <path>` works
- [x] CI release workflow green

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -37,7 +37,7 @@ jobs:
         run: |
           mkdir -p build
           # Use Crystal container to build static binary
-          docker run --rm -v $(pwd):/noir -w /noir --entrypoint="" 84codes/crystal:latest-debian-12 sh -c "
+          docker run --rm -v $(pwd):/noir -w /noir --entrypoint="" crystallang/crystal:1.19.1 sh -c "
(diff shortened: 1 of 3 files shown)


hahwul:
Pushed an update addressing the previous CI failure and the review feedback:

- Switched builder to `crystallang/crystal:1.19.0-alpine` (matches CI's Crystal 1.19.0).
- Dropped the `pkg-config` wrapper hack — alpine's dev packages provide static-linking-ready `.pc` files out of the box.
- Avoided the Ubuntu apt mirrors that flaked the previous run; Alpine `apk` only needs `yaml-dev` and `zstd-dev`.
- Made the AGENTS.md one-liner self-contained so it works against the slimmer official image.

Verified locally on linux/arm64: `docker build .` succeeds and the resulting `noir --version` returns `0.29.1`.

PR #1233, opened 2026-04-19
feat(passive_scan): log regex compilation failures in matchers

### LINKED ISSUE #1092: Passive scan regex compilation failures are silently ignored
## Description

When regex patterns fail to compile in passive scan matchers, the errors are silently swallowed by bare `rescue` blocks.

**File:** `src/models/passive_scan.cr` (lines 38-42, 44-48)

```crystal
begin
  @compiled_regex = Regex.union(@patterns.map { |p| Regex.new(p.to_s) })
rescue
  @compiled_regex = nil
end
```

A debug log message would help users diagnose broken passive scan rules with invalid regex patterns.

tmchow:
## Summary

Resolves #1092. The two bare `rescue` blocks in `Matcher#initialize` (one for the `or` condition / `Regex.union`, one for the `and` condition / per-pattern compile) silently swallow regex compilation errors and fall back to `nil`. With no log output, a malformed regex in a passive scan rule is invisible to operators.

## Changes

`src/models/passive_scan.cr`:
- Add `require "log"` alongside the existing requires.
- Change `rescue` to `rescue ex` in both blocks.
- Emit `Log.debug { ... }` with the error message, error class, and the `@patterns` list before falling back to `nil`.

Fallback behavior is unchanged — `@compiled_regex` / `@compiled_regexes` still get set to `nil` when compilation fails, so existing callers in `src/passive_scan/detect.cr` continue to work.

## Why `Log.debug`

Matches the pattern already used for similar rescue-plus-fallback cases in:
- `src/llm/cache.cr` (cache fetch/store/delete/clear/purge/stats)
- `src/analyzer/analyzers/file_analyzers/graphql_analyzer.cr` (file read error)

Using `Log.debug` (Crystal stdlib) keeps the fix consistent with those call sites and avoids pulling `NoirLogger` into `Matcher#initialize`, which isn't reachable from YAML deserialization.

Closes #1092

### DIFF
--- a/src/models/passive_scan.cr
+++ b/src/models/passive_scan.cr
@@ -39,14 +39,14 @@ struct PassiveScan
           begin
             @compiled_regex = Regex.union(@patterns.map { |p| Regex.new(p.to_s) })
           rescue ex
-            Log.debug { "Passive scan matcher regex compilation (or-union) failed: #{ex.message} (#{ex.class}); patterns=#{@patterns.inspect}" }
+            Log.warn { "Passive scan matcher regex compilation (or-union) failed: #{ex.message} (#{ex.class}); patterns=#{@patterns.map(&.to_s).inspect}" }
             @compiled_regex = nil
           end
         elsif @condition == "and"
           begin
             @compiled_regexes = @patterns.map { |p| Regex.new(p.to_s) }
           rescue ex
-            Log.debug { "Passive scan matcher regex compilation (and-case) failed: #{ex.message} (#{ex.class}); patterns=#{@patterns.inspect}" }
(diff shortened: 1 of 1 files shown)


tmchow:
Update from gemini bot feedback: `debug` does hide this from operators by default, which undercuts the whole reason for surfacing it. Pushed 2225445 that:

1. Elevates both rescue branches from `Log.debug` to `Log.warn`.
2. Uses `@patterns.map(&.to_s).inspect` so the output is clean string patterns rather than the raw `YAML::Any` wrapper.

hahwul (approved):
LGTM

hahwul:
@tmchow Thanks for the PR! Great work, merged.

tmchow:
> @tmchow Thanks for the PR! Great work, merged.

Awesome, thanks!

PR #1235, opened 2026-04-21
refactor(analyzer): 3-layer architecture + PHP/Ruby/Rust engines

hahwul:
## Summary

- Document the analyzer 3-layer model (Language Engine → Route Extractor → Framework Adapter) in `AGENTS.md`, naming Hono as the reference implementation.
- Introduce `PhpEngine`, `RubyEngine`, `RustEngine` under `src/analyzer/engines/` so the 16 analyzers in those languages (previously with no shared base) now inherit worker-pool, file filtering, and error-handling boilerplate from a single place.
- Net **−395 LOC across 16 analyzers** with no material behavior change (see **Notes on behavior** below).

## Layer breakdown

| Layer | Responsibility | New / existing |
|---|---|---|
| Language Engine | file walking, concurrency, per-path error handling | `engines/php_engine.cr`, `ruby_engine.cr`, `rust_engine.cr` (new); Go/Python already had per-language commons |
| Route Extractor | parse source, yield routes | unchanged (JS only; future work) |
| Framework Adapter | apply framework-specific param rules | the analyzers themselves |

## Engine designs

- **PhpEngine** — abstract `analyze_file(path) : Array(Endpoint)`; engine owns channel + `parallel_analyze` + error catch. Shared helpers: `build_full_path`, `extract_brace_path_params`.
- **RubyEngine** — opt-in helpers: `line_to_endpoint(content, details?)` (replaces ~42 duplicated lines between Hanami and Sinatra) and `parallel_file_scan(&block)` for whole-tree scanners. Rails keeps its internal logic; only the superclass changed.
- **RustEngine** — same abstract-`analyze_file` shape as PhpEngine with a `.rs` extension filter baked in. All 9 Rust analyzers had an identical 20-line worker-pool skeleton that now lives in one place.

## Notes on behavior

Two intentional normalizations; neither changes output on real-world inputs but they deviate from the pre-refactor behavior and are worth calling out explicitly:

1. **Concurrency clamp.** Migrated analyzers now go through `Analyzer#parallel_analyze`, which clamps `--concurrency` at `MAX_ANALYZER_WORKERS = 64`. Previously PHP/Ruby/Rust analyzers spawned exactly as many workers as requested. This aligns PHP/Ruby/Rust with Go/JS and is a safety bound, not a throttle anyone will hit in practice (default concurrency is well below 64).
2. **`{id?}` now matches in Symfony.** `extract_brace_path_params` uses `/\{(\w+)\??\}/`, where the original Symfony regex was `/\{(\w+)\}/`. `{id?}` is a Laravel idiom and shouldn't appear in Symfony sources, so the expansion is theoretical. If it ever appears, Symfony will now emit the path param rather than silently dropping it.

Errors that were previously swallowed by bare `rescue` blocks in some Rust analyzers (axum, rocket, rwf) are now logged at debug level via the engine's uniform `rescue e / logger.debug` — strict debuggability improvement.

## Out of scope

- Go/Python commons stay in `analyzers/{lang}/common.cr` since they currently mix engine and route-extraction helpers; they migrate when the engine-vs-extractor split lands.
- No param-extraction unification: semantics differ per framework (Rails emits query+body, Hanami query only, Flask-RESTx differs again). Regex-level consolidation would conflate them.
- Engine-level unit tests are not added in this PR; existing functional tests exercise the engines end-to-end via all 16 migrated analyzers.

## Test plan

- [x] `just build` — clean compile
- [x] `just test` — 1261 unit + 4131 functional examples, 0 failures, 0 errors
- [x] `crystal tool format --check` — all touched files clean

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -75,10 +75,28 @@ spec/
 - `.ameba.yml` - Linting configuration
 - `.github/workflows/ci.yml` - CI configuration
 
+## Analyzer Layering
+
+An analyzer is composed of three layers. Keep them separate — a framework adapter should not open files or re-implement parsing.
+
+1. **Language Engine** — shared per-language base (`src/analyzer/analyzers/{lang}/common.cr` or equivalent). Owns file walking, concurrency, worker pool, file-content caching. One per language.
+2. **Route Extractor** — shared per-language parser layer (`src/miniparsers/{lang}_route_extractor.cr`). Takes source content, yields route declarations (method, path, location). No file I/O, no framework-specific rules.
+3. **Framework Adapter** — thin per-framework class (`src/analyzer/analyzers/{lang}/{framework}.cr`). Consumes routes from the extractor and applies framework-specific param mappings, filters, and special cases.
+
+**Rule**: the framework adapter receives routes; it does not walk the filesystem or parse tokens itself.
+
+**Reference implementation**: `src/analyzer/analyzers/javascript/hono.cr` on top of `src/miniparsers/js_route_extractor.cr`. Hono is ~205 lines because it follows this split; contrast with analyzers that inline all three responsibilities and grow to 500–800 lines.
+
+**Current coverage**:
+- Language engines: Go, Python. Missing for PHP, Ruby, Rust (tracked as separate work).
+- Route extractors: JavaScript only. Python/Kotlin/Java have parsers but not dedicated extractors.
+
+When adding a new framework in a language that already has an extractor, extend the extractor rather than re-parsing inline.
+
 ## Adding New Components
 
 ### Analyzers
-1. Create `src/analyzer/analyzers/{language}/{framework}.cr`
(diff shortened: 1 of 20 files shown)


hahwul:
> ## Code Review
> This pull request refactors the analyzer architecture by introducing language-specific engines for PHP, Ruby, and Rust, centralizing common logic and reducing boilerplate in individual framework adapters. Documentation in AGENTS.md has been updated to reflect this new layering. Critical race conditions were identified in the new engine implementations where shared result arrays are modified concurrently within parallel processing blocks, requiring synchronization with a Mutex.

noir는 preview_mt 안 써서 fiber cooperative라 result << ...는 atomic으로 특이사항 없음. 나중에 MT 커면 그 때 한번에 처리 필요

hahwul:
noir는 `preview_mt` 안 쓰니까 fiber cooperative라 `result << ...`는 atomic. 기존 analyzer 30개도 다 같은 패턴. MT 켜게 되면 `parallel_analyze` 레벨에서 한 번에 처리할 일.

Issue #1366, opened 2026-05-11
Endpoint callees: 1-hop handler call graph for AI SAST context


hahwul:
## Goal

Enrich Noir's endpoint inventory with a per-endpoint list of 1-hop function/method invocations observed inside the handler body. The goal is to give LLM-based code reviewers and AI SAST tools a focused "where to look next" map: names like `User.create`, `raw_sql_query`, or `notify_admin` carry direct signal about what an endpoint touches without making the model read the whole file first.

This is incremental enrichment of the existing context surface (URL, params, headers, tags, path, line), not a standalone feature.

## Design

- New field: `Endpoint.callees : Array(Callee)`, parallel to `tags` and `params`.
- `Callee = { name, path?, line? }`; `path`/`line` are best-effort locations. Most analyzers still report call sites; analyzers with definition resolution report callee definition locations when reachable and fall back to call sites otherwise.
- `--include-callee` toggles plain-text rendering. JSON/YAML/JSONL/TOML serialize the field automatically through the model.
- Per-endpoint cap of 10, deduped on push.
- Builtins (`len`, `print`, `range`, `open`, etc.) and dunder methods are filtered.
- Framework calls (`jsonify`, `render_template`, `request.cookies.get`) are intentionally kept because they show how the endpoint shapes input/output.

## Scope Notes

- 1-hop only. Callees are a useful prior, not a complete dependency graph.
- Dynamic dispatch is out of scope for this pass. Middleware, decorators, Rails `before_action`, Express/Koa middleware, Python `getattr`/`send`-style indirection, and similar patterns may be missing.
- Per-analyzer best effort. Static/named handler frameworks are cleaner targets than heavily inline/lambda-heavy frameworks.

## Status From Code Audit (2026-05-13)

The issue text previously lagged the codebase. Current merged coverage is broader than the original follow-up list: Python, Go, and the first JVM/Kotlin batches are already mostly wired.

### Core

- [x] Model + CLI + Python helper + Flask wiring — #1365
  - `Callee` struct + `Endpoint.callees` + `push_callee` dedup/cap
  - `--include-callee` plain-text rendering
  - `Noir::PythonCalleeExtractor`
  - Functional tester support for callee assertions

### Python

- [x] Flask — #1365
- [x] Sanic, Aiohttp, Bottle, Falcon, Pyramid — #1367
- [x] FastAPI, Litestar — #1368
- [x] Django, Tornado — #1372
- [x] Starlette — #1388

### Go

[...]

PR #1388, opened 2026-05-13
feat(python): wire callees through Starlette

hahwul:
## Summary
- attach 1-hop handler callees to Starlette Route endpoints via the existing Python callee extractor
- reuse Starlette handler resolution for positional and `endpoint=` route declarations
- add a dedicated `starlette_callees` functional fixture/spec covering multi-method routes and Mount prefixes

## Tests
- `crystal spec spec/functional_test/testers/python/starlette_callees_spec.cr spec/functional_test/testers/python/starlette_spec.cr`
- `just test`
- `just build`
- `./bin/noir -b spec/functional_test/fixtures/python/starlette_callees --include-callee`

Refs #1366


hahwul:
Addressed the review comment in a73e376d. Starlette now caches `(params, callees)` per handler inside `analyze_file`, so `parse_code_block` and the callee tree-sitter walk run once per referenced handler and are reused across multi-method routes.

Re-ran:
- `crystal spec spec/functional_test/testers/python/starlette_callees_spec.cr spec/functional_test/testers/python/starlette_spec.cr`
- `just test`
- `just build`
- `./bin/noir -b spec/functional_test/fixtures/python/starlette_callees --include-callee`

/review

PR #1392, opened 2026-05-13
feat(java): wire callees through Play

hahwul:
## Summary
- attach 1-hop controller method callees to Play routes
- parse controller files with tree-sitter and reuse `JavaCalleeExtractor.callees_in_method`
- add a `play_callees` fixture/spec covering route-to-controller callee propagation

## Notes
- Chained calls such as `request().header(...)` follow the existing Java callee policy: outer chained calls are dropped and the inner call (`request`) is kept.

## Tests
- `crystal spec spec/functional_test/testers/java/play_callees_spec.cr spec/functional_test/testers/java/play_spec.cr`
- `just test`
- `just build`
- `./bin/noir -b spec/functional_test/fixtures/java/play_callees --include-callee`

Refs #1366

### DIFF
--- a/spec/functional_test/fixtures/java/play_callees/app/controllers/AppController.java
+++ b/spec/functional_test/fixtures/java/play_callees/app/controllers/AppController.java
@@ -1,0 +1,45 @@
+package controllers;
+
+import play.libs.Json;
+import play.mvc.Controller;
+import play.mvc.Result;
+
+public class AppController extends Controller {
+    public Result create(String id) {
+        String token = request().header("X-Token");
+        User user = parseUser();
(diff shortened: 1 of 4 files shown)


hahwul:
Addressed the review comment in 1cd3e7ba. Play controller discovery now uses the Tree-Sitter AST for class/method traversal, and the same method body node feeds both existing parameter scanning and `JavaCalleeExtractor.callees_in_body`.

Re-ran:
- `crystal spec spec/functional_test/testers/java/play_callees_spec.cr spec/functional_test/testers/java/play_spec.cr`
- `just test`
- `just build`

/review

PR #1393, opened 2026-05-13
feat(java): wire callees through Vert.x

hahwul:
## Summary
- attach Java 1-hop callees to Vert.x routes that use `this::handler` method references
- keep the existing regex route discovery path, then resolve referenced handler method bodies with Tree-sitter
- add a Vert.x callee fixture and functional regression spec

## Verification
- `crystal spec spec/functional_test/testers/java/vertx_callees_spec.cr spec/functional_test/testers/java/vertx_spec.cr`
- `just test`
- `just build`
- `just check`
- `./bin/noir -b spec/functional_test/fixtures/java/vertx_callees --include-callee`

Refs #1366

### DIFF
--- a/spec/functional_test/fixtures/java/vertx_callees/src/main/java/com/example/MainVerticle.java
+++ b/spec/functional_test/fixtures/java/vertx_callees/src/main/java/com/example/MainVerticle.java
@@ -1,0 +1,61 @@
+package com.example;
+
+import io.vertx.core.Vertx;
+import io.vertx.ext.web.Router;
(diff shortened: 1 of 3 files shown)


hahwul:
Addressed the review suggestion: `walk_method_declarations` now returns immediately after yielding a `method_declaration`, so it no longer descends into method bodies while looking for handler declarations.

Re-verified locally:
- `crystal spec spec/functional_test/testers/java/vertx_callees_spec.cr spec/functional_test/testers/java/vertx_spec.cr`
- `just check`
- `just build`

/review

PR #1401, opened 2026-05-13
feat(js): wire callees through SvelteKit

hahwul:
## Summary
- attach shared JS callee extraction to SvelteKit `+server` route handlers when `include_callee` is enabled
- scope callees to explicit verb exports and same-file export aliases without attaching page-component or fallback-route callees
- add focused SvelteKit callee fixture coverage for typed `RequestHandler`, function exports, and alias exports

## Tests
- `crystal spec spec/functional_test/testers/javascript/sveltekit_callees_spec.cr spec/functional_test/testers/javascript/sveltekit_spec.cr`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-sveltekit just build`
- `./bin/noir -b spec/functional_test/fixtures/javascript/sveltekit_callees --include-callee`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-sveltekit just check`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-sveltekit just test`

### DIFF
--- a/spec/functional_test/fixtures/javascript/sveltekit_callees/package.json
+++ b/spec/functional_test/fixtures/javascript/sveltekit_callees/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "sveltekit-callees-fixture",
+  "private": true,
+  "dependencies": {
+    "@sveltejs/kit": "^2.0.0",
+    "@sveltejs/adapter-node": "^5.0.0"
+  }
+}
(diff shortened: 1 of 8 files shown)


hahwul:
Addressed the review feedback in 66378a54. SvelteKit no longer has analyzer-local regex/brace body extraction. Handler lookup now uses a shared AST-based `JSCalleeExtractor.callees_for_exported_function`, so line numbers come from Tree-Sitter rows and implicit-return arrow handlers are covered. Added unit coverage for braces in strings/comments, alias exports, implicit returns, and typed `RequestHandler` exports.

/review

PR #1409, opened 2026-05-13
Add Hanami callee extraction

hahwul:
## Summary
- attach Ruby callee extraction to Hanami action `handle` bodies when `--include-callee` is enabled
- keep Hanami param extraction behavior while reusing the same action file read for callee extraction
- cover multiline, inline, endless-method, missing-action, params DSL, and nested branch Hanami actions

## Tests
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-hanami crystal spec spec/unit_test/miniparser/ruby_callee_extractor_spec.cr spec/functional_test/testers/ruby/hanami_spec.cr spec/functional_test/testers/ruby/hanami_callees_spec.cr`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-hanami just build`
- `./bin/noir -b spec/functional_test/fixtures/ruby/hanami_callees --include-callee`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-hanami just check`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-hanami just test`

Part of #1366

### DIFF
--- a/spec/functional_test/fixtures/ruby/hanami_callees/Gemfile
+++ b/spec/functional_test/fixtures/ruby/hanami_callees/Gemfile
@@ -1,0 +1,3 @@
+source "https://rubygems.org"
+
+gem "hanami", "~> 2.0"
--- a/spec/functional_test/fixtures/ruby/hanami_callees/app/action.rb
+++ b/spec/functional_test/fixtures/ruby/hanami_callees/app/action.rb
@@ -1,0 +1,6 @@
+require "hanami/action"
+
+module Testapp
+  class Action < Hanami::Action
+  end
+end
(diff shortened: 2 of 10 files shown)


hahwul:
Gemini Code Assist could not run because the daily quota was exhausted. I reviewed this locally with parallel read-only agent checks instead and addressed the findings before opening the PR:

- inline `def handle(request, response); ...; end` extraction now skips the signature args before reading the body
- Ruby endless method `def handle(...) = ...` is covered by the fixture
- the POST fixture now separates JSON params from request-derived query params (`source`) so expected params are explicit

Local and GitHub Crystal build/test/lint checks are passing. Docker is intentionally not blocking this merge.

PR #1423, opened 2026-05-13
Add Grip callee extraction

hahwul:
## Summary
- wire Grip controller routes to attach 1-hop Crystal callees when `--include-callee` is enabled
- resolve `as: :action` route targets and default HTTP verb method targets
- resolve websocket routes to `on_message` by default
- preserve existing endpoint/param semantics while adding a focused callee fixture

## Scope notes
- This resolves same-file Grip controller methods only.
- Handler inference is intentionally limited to existing Grip route forms and websocket `on_message`.
- Marten remains separate because its current analyzer emits one `GET` per `path` and needs careful handler mapping.

## Tests
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-grip-target-a crystal spec spec/functional_test/testers/crystal/grip_spec.cr spec/functional_test/testers/crystal/grip_callees_spec.cr`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-grip-build just build`
- `./bin/noir -b spec/functional_test/fixtures/crystal/grip_callees --only-techs crystal_grip --include-callee`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-grip-check just check`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-grip-test just test`

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/crystal/grip_callees/shard.yml
+++ b/spec/functional_test/fixtures/crystal/grip_callees/shard.yml
@@ -1,0 +1,12 @@
+name: grip_callees
+version: 0.1.0
+
+targets:
+  grip_callees:
+    main: src/testapp.cr
+
+dependencies:
+  grip:
+    github: grip-framework/grip
+
+crystal: 1.8.2
--- a/spec/functional_test/fixtures/crystal/grip_callees/src/testapp.cr
+++ b/spec/functional_test/fixtures/crystal/grip_callees/src/testapp.cr
@@ -1,0 +1,61 @@
+require "grip"
+
+class IndexController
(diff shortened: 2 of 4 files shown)


hahwul:
Addressed the review follow-up.

- Replaced the hardcoded class-depth tracker with a module/class/block stack.
- Namespaced controllers such as `Api::StatusController` are now collected as fully-qualified action keys.
- Added a Grip callee fixture assertion for a module-nested controller route.
- Re-ran targeted Grip specs, build, check, CLI smoke, and full test locally.

/review

PR #1424, opened 2026-05-13
Add Marten callee extraction

hahwul:
## Summary
- add `--include-callee` support for Marten `path "...", Handler` routes
- resolve handler classes across project files and attach 1-hop callees from the matched handler `def get` body
- support namespaced handlers such as `Admin::ReportsHandler`
- add focused Marten callee fixture coverage, including a `post` method that must not bleed into the emitted GET route

## Scope notes
- Existing Marten endpoint semantics are preserved: each `path` declaration still emits a GET endpoint only.
- This PR does not fan out POST/PUT/DELETE endpoints from handler methods.
- Existing raw empty-url handler-method detections are left unchanged; optimizer still reduces the fixture to the expected final route endpoints.
- Callee resolution is static and keyed by handler class name plus `get`.

## Tests
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-marten-target crystal spec spec/functional_test/testers/crystal/marten_spec.cr spec/functional_test/testers/crystal/marten_callees_spec.cr`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-marten-build just build`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-marten-check just check`
- `./bin/noir -b spec/functional_test/fixtures/crystal/marten_callees --only-techs crystal_marten --include-callee`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-marten-test just test`

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/crystal/marten_callees/shard.yml
+++ b/spec/functional_test/fixtures/crystal/marten_callees/shard.yml
@@ -1,0 +1,18 @@
+name: testmarten
+version: 0.1.0
+
+authors:
+  - example <example@gmail.com>
+
+targets:
+  testmarten:
+    main: src/testmarten.cr
+
+dependencies:
+  marten:
+    github: martenframework/marten
+    version: ~> 0.4.0
+
+crystal: 1.8.2
+
+license: MIT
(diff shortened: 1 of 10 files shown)


hahwul:
Addressed the review follow-up.

- Guarded module/class scope pushes for single-line declarations such as `class PlaceholderHandler; end`.
- Normalized absolute constants such as `::Admin::ReportsHandler` so they resolve to the same handler key.
- Made handler class lookup search back through class/block scope, with fixture coverage for a class-level block wrapping `def get`.
- Re-ran targeted Marten specs, build, check, CLI smoke, and full test locally.

/review

PR #1431, opened 2026-05-14
Add Tide callee extraction

hahwul:
## Summary
- wire Tide named route handlers to attach Rust callees when include_callee is enabled
- reuse shared Rust function-body slicing and RustCalleeExtractor for same-file async fn handlers
- preserve existing Tide path/query/json/form/header/cookie param extraction
- add a focused Tide callee fixture covering direct app.at routes, route variable assignment, and path/query/json/header/cookie params

## Scope notes
- Supports same-file named async fn handlers referenced by Tide route registrations.
- Inline closure handlers keep existing endpoint behavior but are out of callee scope for this pass.
- Existing Tide route location metadata remains unchanged.

## Tests
- CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-tide-target crystal spec spec/functional_test/testers/rust/tide_spec.cr spec/functional_test/testers/rust/tide_callees_spec.cr
- CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-tide-build just build
- CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-tide-check just check
- CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-tide-test just test
- ./bin/noir -b spec/functional_test/fixtures/rust/tide_callees -f json --include-callee

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/rust/tide_callees/Cargo.toml
+++ b/spec/functional_test/fixtures/rust/tide_callees/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "tide-callees-example"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+tide = "0.16"
+async-std = { version = "1", features = ["attributes"] }
+serde = { version = "1", features = ["derive"] }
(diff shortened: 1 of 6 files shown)


hahwul:
Addressed review feedback in b7b8e035:

- unified Tide handler function scanning so parameter bodies and callees are collected in the same pass
- exposed comment-state stripping and updated the Rust function-body helper to return the consumed end line
- skipped consumed function bodies during Tide scanning and added block-comment regression coverage
- tightened Tide handler-name matching to full identifiers/paths only

Re-running review now.

/review

PR #1432, opened 2026-05-14
Add Gotham callee extraction

hahwul:
## Summary
- wire Gotham route handlers to attach Rust callees when --include-callee is enabled
- reuse the shared Rust function-body helper for handler param extraction and callee extraction
- keep existing Gotham cookie/header/path param extraction while adding callee fixture coverage

## Scope
- supports same-file named Gotham handlers passed through .to(handler) or .to(module::handler)
- keeps call-site path/line behavior consistent with the other Rust callee analyzers

## Tests
- crystal spec spec/functional_test/testers/rust/gotham_spec.cr spec/functional_test/testers/rust/gotham_callees_spec.cr
- just build
- just check
- just test
- ./bin/noir -b spec/functional_test/fixtures/rust/gotham_callees -f json --include-callee

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/rust/gotham_callees/Cargo.toml
+++ b/spec/functional_test/fixtures/rust/gotham_callees/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "gotham-callees-example"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+gotham = "0.7"
+hyper = "0.14"
+tokio = { version = "1", features = ["full"] }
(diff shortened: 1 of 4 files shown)


hahwul:
Addressed review feedback in 4d406e96:

- precomputed Gotham handler function bodies once per file instead of scanning the whole file per route
- added comment-stripping that preserves string literals for cookie/header parameter extraction
- moved cookie/header regex matching to stripped code lines to avoid comment false positives
- added fixture coverage for inline block comments before real cookie/header calls

Re-running review now.

/review

PR #1433, opened 2026-05-14
Add Loco callee extraction

hahwul:
## Summary
- attach Rust callees for Loco endpoints when --include-callee is enabled
- reuse the shared Rust function-body and callee extraction helpers for each inferred handler action
- add Loco controller fixture coverage for REST-style actions, headers, form data, and JSON bodies

## Scope
- same-file handler body only; Loco's existing action-name/path inference remains unchanged
- keeps existing Loco parameter extraction behavior

## Tests
- crystal spec spec/functional_test/testers/rust/loco_spec.cr spec/functional_test/testers/rust/loco_callees_spec.cr
- just build
- just check
- just test
- ./bin/noir -b spec/functional_test/fixtures/rust/loco_callees -f json --include-callee

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/rust/loco_callees/Cargo.toml
+++ b/spec/functional_test/fixtures/rust/loco_callees/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "loco-callees-example"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+loco-rs = { version = "0.2" }
+tokio = { version = "1", features = ["full"] }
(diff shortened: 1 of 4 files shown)


hahwul:
Addressed review feedback in 6c033c77:

- extract each Loco handler body once and reuse it for both parameter body scans and callee extraction
- removed the duplicate body extraction path when --include-callee is enabled
- parameter scans now use the full extracted function body when available instead of the previous 30-line body scan fallback

Re-running review now.

/review

PR #1434, opened 2026-05-14
Add RWF callee extraction

hahwul:
## Summary
- attach Rust callees for RWF controller handle bodies when --include-callee is enabled
- precompute controller handle bodies once per file and reuse them for method, parameter, and callee extraction
- add RWF callee fixture coverage for method dispatch, path/header/query/cookie/body params, and comment-safe header extraction

## Scope
- callee extraction is scoped to each controller handle body; when one handle dispatches multiple HTTP methods, emitted endpoints share the controller-level 1-hop callees
- preserves existing RWF route! path and method inference behavior

## Tests
- crystal spec spec/functional_test/testers/rust/rwf_spec.cr spec/functional_test/testers/rust/rwf_callees_spec.cr
- just build
- just check
- just test
- ./bin/noir -b spec/functional_test/fixtures/rust/rwf_callees -f json --include-callee

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/rust/rwf_callees/Cargo.toml
+++ b/spec/functional_test/fixtures/rust/rwf_callees/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "rwf-callees-example"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+rwf = "0.2"
+tokio = { version = "1", features = ["full"] }
+serde_json = "1"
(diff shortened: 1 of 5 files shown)


hahwul:
Addressed review feedback in a1d25e07:

- switched RWF header/cookie extraction to the existing typed parameter helper
- added a preserve_strings option to RustCalleeExtractor's comment stripping helper
- removed the RWF-local comment stripping duplicate and reused the shared helper for parameter scans

Re-running review now.

/review

PR #1435, opened 2026-05-14
Add Warp callee extraction

hahwul:
## Summary
- add callee extraction for named Warp handlers passed to .map, .and_then, and .then
- collect same-file Rust handler function callees and attach them to matching endpoints when --include-callee is enabled
- add Warp callee fixtures and functional coverage for GET, POST, path, query, and JSON params

## Scope
- covers named handlers in the same source file
- inline closures keep existing endpoint/parameter behavior but do not emit callees yet

## Verification
- crystal spec spec/functional_test/testers/rust/warp_spec.cr spec/functional_test/testers/rust/warp_callees_spec.cr
- just build
- just check
- just test
- ./bin/noir -b spec/functional_test/fixtures/rust/warp_callees -f json --include-callee

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/rust/warp_callees/Cargo.toml
+++ b/spec/functional_test/fixtures/rust/warp_callees/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "warp-callees-example"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+warp = "0.3"
+tokio = { version = "1", features = ["full"] }
+serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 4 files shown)


hahwul:
Addressed review feedback in 3acd0a50:

- expanded Warp function discovery to accept repeated async/const/unsafe modifiers after visibility
- added turbofish support for named handlers passed to .map/.and_then/.then
- added a functional regression route for a namespaced turbofish handler

Verification rerun:
- crystal spec spec/functional_test/testers/rust/warp_spec.cr spec/functional_test/testers/rust/warp_callees_spec.cr
- just build
- just check
- just test
- ./bin/noir -b spec/functional_test/fixtures/rust/warp_callees -f json --include-callee

/review

PR #1437, opened 2026-05-14
Add Elixir Phoenix callee extraction

hahwul:
## Summary
- attach Elixir Phoenix controller action callees to route endpoints when `--include-callee` is enabled
- reuse the shared Elixir callee extractor added for Plug
- add Phoenix controller/route fixtures and functional coverage for direct routes and `resources` routes

## Scope
- covers controller action bodies mapped through Phoenix router/controller/action entries
- LiveView/socket routes are still endpoint-only because there is no controller action body to attach

## Verification
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-phoenix crystal spec spec/unit_test/miniparser/elixir_callee_extractor_spec.cr spec/functional_test/testers/elixir/phoenix_spec.cr spec/functional_test/testers/elixir/phoenix_callees_spec.cr`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-phoenix-check just check`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-phoenix-build just build`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-phoenix-test just test`
- `./bin/noir -b spec/functional_test/fixtures/elixir/phoenix_callees -f json --include-callee`

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/elixir/phoenix_callees/lib/elixir_phoenix_callees_web/controllers/post_controller.ex
+++ b/spec/functional_test/fixtures/elixir/phoenix_callees/lib/elixir_phoenix_callees_web/controllers/post_controller.ex
@@ -1,0 +1,15 @@
+defmodule ElixirPhoenixCalleesWeb.PostController do
+  use ElixirPhoenixCalleesWeb, :controller
+
+  def index(conn, _params) do
+    category = conn.query_params["category"]
+    posts = PostQuery.list(category)
+    render(conn, :index, posts: PostPresenter.render(posts))
+  end
+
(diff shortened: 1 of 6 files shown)


hahwul:
Addressed the review note in `9da0517a`.

- Phoenix now collects matching endpoints per controller action before doing block work.
- The action block end and callee list are computed once per function definition, then reused for each matching endpoint.
- Re-verified with targeted Phoenix/Elixir callee specs, `just build`, `just check`, and `git diff --check`.

PR #1438, opened 2026-05-14
Add Swift Vapor callee extraction

hahwul:
## Summary
- add a lightweight shared Swift callee extractor for 1-hop receiver and bare call sites
- attach Vapor route closure callees when `--include-callee` is enabled
- add Vapor callee fixtures and functional coverage for multiline and single-line route closures

## Scope
- covers inline Vapor route closure bodies in the existing Vapor analyzer path
- keeps existing Vapor endpoint and parameter extraction semantics unchanged
- extractor is best-effort regex-based and skips comments/strings, not a full Swift AST call graph

## Verification
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-vapor-target crystal spec spec/unit_test/miniparser/swift_callee_extractor_spec.cr spec/functional_test/testers/swift/vapor_callees_spec.cr spec/functional_test/testers/swift/vapor_spec.cr`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-vapor-build just build`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-vapor-check just check`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-vapor-test just test`
- `./bin/noir -b spec/functional_test/fixtures/swift/vapor_callees -f json --include-callee`

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/swift/vapor_callees/Package.swift
+++ b/spec/functional_test/fixtures/swift/vapor_callees/Package.swift
@@ -1,0 +1,14 @@
+// swift-tools-version:5.9
+import PackageDescription
+
+let package = Package(
+    name: "VaporCalleesFixture",
+    dependencies: [
+        .package(url: "https://github.com/vapor/vapor.git", from: "4.0.0"),
+    ],
+    targets: [
+        .target(name: "App", dependencies: [
+            .product(name: "Vapor", package: "vapor"),
+        ]),
+    ]
+)
--- a/spec/functional_test/fixtures/swift/vapor_callees/routes.swift
+++ b/spec/functional_test/fixtures/swift/vapor_callees/routes.swift
@@ -1,0 +1,20 @@
+import Vapor
+
+func routes(_ app: Application) throws {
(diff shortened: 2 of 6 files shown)


hahwul:
Addressed the review notes in `ca29fbf8`.

- Preserved route-body line alignment by keeping the same-line fragment even when it is empty.
- Reused `SwiftCalleeExtractor.strip_non_code_with_state` for Vapor route-body brace counting so braces in comments and strings do not close the route body early.
- Added nested Swift block-comment depth handling and multiline string tracking in the shared extractor.
- Added bare trailing-closure call extraction while filtering common control-flow conditions such as `if shouldAudit {`.
- Added unit/functional coverage for nested comments, multiline strings, trailing closure calls, and route bodies containing `}` noise.

Re-verified with targeted Swift specs, `just build`, `just check`, `just test`, `git diff --check`, and the Vapor callee CLI smoke.

PR #1439, opened 2026-05-14
Add Swift Hummingbird callee extraction

hahwul:
## Summary
- attach Hummingbird route closure callees when `--include-callee` is enabled
- reuse the shared Swift callee extractor introduced for Vapor
- add Hummingbird callee fixtures and functional coverage for body/path/query params, multiline closures, and single-line closures

## Scope
- covers inline Hummingbird route closure bodies in the existing analyzer path
- keeps existing Hummingbird endpoint and parameter extraction semantics unchanged
- route-body slicing uses the shared Swift non-code stripper so braces inside comments/strings do not end the body early

## Verification
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-hummingbird-target crystal spec spec/unit_test/miniparser/swift_callee_extractor_spec.cr spec/functional_test/testers/swift/hummingbird_callees_spec.cr spec/functional_test/testers/swift/hummingbird_spec.cr`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-hummingbird-build just build`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-hummingbird-check just check`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-hummingbird-test just test`
- `./bin/noir -b spec/functional_test/fixtures/swift/hummingbird_callees -f json --include-callee`

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/swift/hummingbird_callees/Package.swift
+++ b/spec/functional_test/fixtures/swift/hummingbird_callees/Package.swift
@@ -1,0 +1,20 @@
+// swift-tools-version:5.9
+import PackageDescription
+
+let package = Package(
+    name: "HummingbirdExample",
+    platforms: [
+       .macOS(.v14)
+    ],
+    dependencies: [
+        .package(url: "https://github.com/hummingbird-project/hummingbird.git", from: "2.0.0"),
+    ],
+    targets: [
+        .executableTarget(
+            name: "App",
+            dependencies: [
+                .product(name: "Hummingbird", package: "hummingbird")
+            ]
+        ),
+    ]
+)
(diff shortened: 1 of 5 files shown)


hahwul:
Addressed the review notes in `dc31f98a`.

- The Swift non-code stripper now preserves original string length with spaces, so stripped-code brace indexes can be safely applied to the original line.
- Hummingbird route body extraction now looks ahead for closures whose opening `{` starts on a following line.
- Closing-brace handling now slices at the matching structural brace instead of appending trailing source after the route body.
- Added fixture coverage for a next-line Hummingbird closure and closing-line outside-call noise.

Re-verified with targeted Hummingbird/Vapor Swift specs, `just build`, `just check`, `just test`, `git diff --check`, and the Hummingbird callee CLI smoke.

PR #1440, opened 2026-05-14
Add Swift Kitura callee extraction

hahwul:
## Summary
- attach Kitura inline route closure callees when `--include-callee` is enabled
- reuse the shared Swift callee extractor for Kitura route bodies
- add same-file bare named handler support for `handler: showProfile`
- add Kitura callee fixtures covering path/body/query params, single-line closures, next-line closures, and named handlers

## Scope
- covers inline route closures and same-file bare `handler: functionName`
- does not resolve receiver-based handlers like `handler: controller.showProfile`
- preserves existing endpoint/param behavior while fixing closure-block termination for `in` at line end in the focused Kitura path

## Verification
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-kitura-target2 crystal spec spec/unit_test/miniparser/swift_callee_extractor_spec.cr spec/functional_test/testers/swift/kitura_callees_spec.cr spec/functional_test/testers/swift/kitura_spec.cr`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-kitura-build2 just build`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-kitura-check2 just check`
- `CRYSTAL_CACHE_DIR=/tmp/noir-crystal-cache-swift-kitura-test2 just test`
- `./bin/noir -b spec/functional_test/fixtures/swift/kitura_callees -f json --include-callee`

Part of #1366.

### DIFF
--- a/spec/functional_test/fixtures/swift/kitura_callees/Package.swift
+++ b/spec/functional_test/fixtures/swift/kitura_callees/Package.swift
@@ -1,0 +1,17 @@
+// swift-tools-version:5.8
+import PackageDescription
+
+let package = Package(
+    name: "KituraCalleesFixture",
+    dependencies: [
+        .package(url: "https://github.com/Kitura/Kitura.git", from: "2.9.0"),
+    ],
+    targets: [
+        .executableTarget(
+            name: "App",
+            dependencies: [
+                .product(name: "Kitura", package: "Kitura")
+            ]
+        ),
+    ]
+)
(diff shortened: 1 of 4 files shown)


hahwul:
Addressed the review feedback in 3a66fb39.\n\n- increased Kitura route body lookahead to the existing 20-line limit\n- indexed same-file named handler bodies once per file instead of rescanning per route\n- match handler function signatures from comment/string-stripped code\n- no longer stop handler body lookup on blank lines between a function signature and opening brace\n- added fixture coverage for commented-out same-name handlers and blank lines before the handler body\n\n/review

PR #1750, opened 2026-05-27
fix: resolve concurrent fiber scheduler crashes and spinner deadlocks

hahwul:
- Refactor parallel_analyze to accept a files list and spawn both the producer and workers inside a single, tracked WaitGroup block.
- Update 9 language engines and custom framework adapters (Go, Java, Python, C++, Dart) to use the new tracked files-based parallel_analyze.
- Remove obsolete untracked channel-based parallel scan APIs and unsafe file helper routines.
- Pre-build CodeLocator extension index synchronously on the main thread before starting parallel analyzers.
- Secure CodeLocator#build_extension_index using a Mutex spinlock to prevent concurrent mutation race conditions under multi-threading.
- Resolve logger spinner fiber crashes and deadlocks by replacing the cooperative Mutex within the raw OS thread with an atomic spinlock (@stdout_busy) and non-blocking skipped frames, completely decoupling spinner rendering from the fiber scheduler.

### DIFF
--- a/src/analyzer/analyzer.cr
+++ b/src/analyzer/analyzer.cr
@@ -254,6 +254,9 @@ def analysis_endpoints(options : Hash(String, YAML::Any), techs, logger : NoirLo
   selected_techs = filter_redundant_generic_techs(techs).select { |t| analyzer.has_key?(t) }
   mutex = Mutex.new
 
+  # Pre-build extension index synchronously to avoid concurrent mutation in multiple threads/fibers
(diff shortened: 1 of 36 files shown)


hahwul:
- 정리하자면 비동기 Fiber 추적을 단일화하고 스피너 스레드 간섭 배제하고자 Atomic Spin Lock 적용
- 결과는 동시 스캔 시 발생하던 크래시와 스피너 데드락을 해결
- 테스트 시 특이사항 없음

PR #1823, opened 2026-05-31
fix(nextjs): filter out commented code

### LINKED ISSUE #1791: JS - Commented-out code detection
**Describe the bug**
Lines starting with `//` (and `/* */` blocks) still get analyzed, so commented-out handlers/headers show up as real endpoints. Feels like it belongs in a shared JS/TS lexer since other analyzers might hit the same thing.

chinmaypant21:
closes #1791

Added nextjs specific test-cases


hahwul (approved):

hahwul:
@chinmaypant21 
Thanks for the PR, and welcome aboard.

I looked into this and the commented-endpoint issue from #1791 turned out to be already fixed by #1777, which added strip_js_comments and runs it over the content used for method detection. When you first branched this was very likely a real bug, but #1777 landed on main in the meantime and quietly resolved it. So this is just a timing overlap, nothing wrong on your end.

The regression test you added is genuinely worth keeping though, since it locks the behavior in and will catch us if the comment stripping ever changes. So I pushed a small cleanup to your branch: dropped the nextjs.cr change (it re-stripped content that was already stripped a line later), restored the fixture indentation, and added a /* */ block-comment case so the test matches its description. All Next.js specs pass.

Appreciate you taking the time on this, and I hope it's the first of many :)

chinmaypant21:
Thanks for the feedback @hahwul.
Quite a coincidence that both the merges were almost on the same time, but ultimately getting the issue fixed is the important part, which was nicely handled by your changes.

The block comment case is also a nice addition that I missed, thank you for that 😄.

PR #1865, opened 2026-06-04
feat(gin): propagate call-site group prefix into router-builder helpers

hahwul:
Stacked on #1864 → #1863 (base retargets to `main` as those merge).

## Problem

The canonical gin layout (used by gin's own `examples/group-routes`) splits registration across `func addXRoutes(rg *gin.RouterGroup)` helpers invoked with a versioned group:

```go
v1 := router.Group("/v1")
addUserRoutes(v1)   // func addUserRoutes(rg *gin.RouterGroup) { users := rg.Group("/users"); users.GET("/", h) }
addPingRoutes(v1)
v2 := router.Group("/v2")
addPingRoutes(v2)
```

The prefix lives at the **call site** and the helper's parameter name (`rg`) does not match the caller's group variable (`v1`). So routes came out as `/users/` instead of `/v1/users/`, and the second `addPingRoutes(v2)` call's `/v2/ping` was missed entirely.

## Fix

New extractor helpers `collect_router_group_builders` / `collect_router_builder_callsites` / `extract_routes_from_function`, plus gin.cr `resolve_router_builder_prefixes`, resolve each builder's call-site prefixes and emit one route set per prefix (so a helper reused under `/v1` and `/v2` yields both).

## Safety (no regression on existing apps)

Only **"case b"** helpers are expanded — those whose group parameter name is NOT already a key in the package group map. **"Case a"** helpers whose parameter happens to match the caller's group variable (e.g. both named `v1`, as in **go-admin** and the existing `gin_engine_param_collision` fixture) are left to the established name-coincidence path, keeping their params/callees. Expanded builder bodies are line-range-suppressed in the whole-file pass to avoid emitting the prefix-less variant alongside the corrected one.

## Validation
- gin-gonic/examples `group-routes`: 4 prefix-less → 5 correct (`/v1/users/`, `/v1/users/comments`, `/v1/users/pictures`, `/v1/ping/`, `/v2/ping/`).
- **go-admin: 175 → 175** (case-a untouched); gin-vue-admin unchanged by this PR; sponge (indirect registry) safely not-expanded, no FP.
- New `node_end_row` tree-sitter helper, `gin_router_builder` fixture + unit test (fail on base); full suite green (functional 16481, unit 2906).

### Known follow-ups (deliberately out of scope)
Inline-group-arg call sites `addX(router.Group("/v1"))` (only group-var args resolved); indirect registries (`[]func(*gin.RouterGroup)` + generic register, e.g. sponge); expanded-route params/callees (matches the chi Mount-expansion precedent).

### DIFF
--- a/spec/functional_test/fixtures/go/gin_router_builder/go.mod
+++ b/spec/functional_test/fixtures/go/gin_router_builder/go.mod
@@ -1,0 +1,5 @@
+module go
+
+go 1.21
+
+require github.com/gin-gonic/gin v1.9.1
--- a/spec/functional_test/fixtures/go/gin_router_builder/main.go
+++ b/spec/functional_test/fixtures/go/gin_router_builder/main.go
@@ -1,0 +1,16 @@
+package routes
+
+import "github.com/gin-gonic/gin"
+
+var router = gin.Default()
+
+// getRoutes splits registration across per-resource helpers, each
+// receiving a versioned group. addPingRoutes is reused under /v1 and /v2.
+func getRoutes() {
+	v1 := router.Group("/v1")
+	addUserRoutes(v1)
+	addPingRoutes(v1)
+
+	v2 := router.Group("/v2")
+	addPingRoutes(v2)
+}
--- a/spec/functional_test/fixtures/go/gin_router_builder/ping.go
+++ b/spec/functional_test/fixtures/go/gin_router_builder/ping.go
@@ -1,0 +1,12 @@
+package routes
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+)
+
+func addPingRoutes(rg *gin.RouterGroup) {
+	ping := rg.Group("/ping")
+	ping.GET("/", func(c *gin.Context) { c.JSON(http.StatusOK, "pong") })
+}
--- a/spec/functional_test/fixtures/go/gin_router_builder/users.go
+++ b/spec/functional_test/fixtures/go/gin_router_builder/users.go
@@ -1,0 +1,15 @@
+package routes
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+)
+
+func addUserRoutes(rg *gin.RouterGroup) {
+	users := rg.Group("/users")
+
+	users.GET("/", func(c *gin.Context) { c.JSON(http.StatusOK, "users") })
+	users.GET("/comments", func(c *gin.Context) { c.JSON(http.StatusOK, "comments") })
+	users.POST("/", func(c *gin.Context) { c.JSON(http.StatusOK, "create") })
+}
--- a/spec/functional_test/testers/go/gin_router_builder_spec.cr
+++ b/spec/functional_test/testers/go/gin_router_builder_spec.cr
@@ -1,0 +1,21 @@
+require "../../func_spec.cr"
+
+# The canonical gin layout splits route registration across
+# `func addXRoutes(rg *gin.RouterGroup)` helpers called from a central
+# function with a versioned group. The prefix (`/v1`) lives at the call
+# site, not in the helper, and the helper's parameter name (`rg`) does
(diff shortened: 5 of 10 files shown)


hahwul:
Converting to **draft** — an adversarial review (and a self-repro) surfaced regressions that need addressing before this is merge-ready. The core idea (graft call-site prefix onto router-builder helpers) is sound and validated on gin's `group-routes` example, but the suppress-then-expand mechanism has gaps:

**Must fix (true regressions vs. pre-PR behavior)**
- **Mixed-mount route loss**: a builder called with BOTH a resolvable group (`addX(v1)`) AND an unresolved receiver (root `*gin.Engine`, or any non-group arg) loses the unresolved call's routes — the body is suppressed but only the resolved prefix is expanded. Guard: only expand when *all* call sites resolve (else fall back to the whole-file pass).
- **params/callees lost on expanded routes**: suppression skips the line-loop, so `c.Query()`/`c.BindJSON()`/cookie params and callees on expanded builder routes are dropped. (Path params like `:id` still come from the optimizer, and default scans don't compute callees, so the practical hit is narrow — but it's real for verbose/inline-handler cases.) Fix likely needs the chi-style `attach_router_function_params` approach over the builder body.

**Should fix (incompleteness, not a regression)**
- **inline-group arg**: `addX(router.Group("/v1"))` isn't resolved (only bare-identifier args like `addX(v1)`). Resolve a `.Group("literal")` first-arg in `collect_router_builder_callsites`.
- minor: `func F(a, b *gin.RouterGroup)` (Go groups same-typed params into one `parameter_declaration`) isn't counted as 2 → would be treated as a single-param builder; whole-line suppression granularity; case-a name-coincidence wrong-prefix edge.

Depends on #1864 (stacked); will rebase onto the updated chi branch when reworking. #1863 and #1864 (zap FP + chi MethodFunc/Mount-resource) are independent and ready.

hahwul:
**Continued implementation + quality pass (after Claude token limit stop + rebase onto main post-#1866).**

- Rebases cleanly on current main (chi Mount/MethodFunc polish included; resolved 3 conflict files by preferring the updated chi logic per the stacked intent in the original draft comment).
- Addressed *all* "must fix" + the inline "should fix" from the self-review comment:
  - Mixed call-site guard: only expand/suppress a builder when *every* callsite resolves to a known prefix (else fall back to whole-file TS pass to avoid losing unresolved routes). Implemented via `__unresolved__` marker in `collect_router_builder_callsites` + reject in `resolve_router_builder_prefixes`.
  - Inline `addX(router.Group("/v1"))`: now supported in collect (detects Group call + string_expr_text on its arg; resolve treats leading-/ arg as direct prefix).
  - Params (and now callees) on expanded: `attach_gin_builder_params` + `add_*_to_many` (modeled on chi `attach_router_function_params` + brace state) so Query/BindJSON/Cookie/Param inside helpers are attached to each prefixed route instance. Callees also wired in the emit (parity with ts_hits path).
- Added unit coverage for inline Group collect path.
- Extracted/factored gin param patterns; brace scoping added to attach to prevent out-of-handler attachments (reviewer feedback).
- All local: `just check` (0 failures on 1300 files), `shards build`, `crystal spec` on miniparser unit + all go/ functional (1452 ex, 0 fail), full unit_test (2911 ex) green.
- CI: core jobs (lint, tests 1.19/1.20, build-crystal) passed on previous push; delta is small + verified.
- Full /review subagent run (thorough read of gin.cr, extractor helpers, chi precedent, fixtures, specs; 50+ tool calls) produced structured notes (positive on guard/inline/param core + tests; 2 bugs noted which are now fixed in last commit).

The feature (prefix grafting for the canonical gin group-routes helper layout, including reuse under multiple versions) is complete and tested. Draft regressions resolved. Ready for merge.

(Review notes artifact from subagent: `/tmp/grok-review-36c49ae0.md` in the session that performed the review.)

PR #1900, opened 2026-06-05
fix(python/flask): strip comments before relevance checks in flask_relevant_source?

hahwul:
The guard added in 0a16014 ("don't claim foreign-framework handler files") used a loose `source.includes?("flask")` *before* the competing-import check. Any file whose *comment* mentioned "flask" (including the new `flask_foreign` fixture's own explanatory comments, and real-world docs or "similar to flask" notes) would be treated as Flask-relevant.

This caused the Flask analyzer to also emit the FastAPI (etc.) routes. Since all Python analyzers run concurrently and results are concatenated before dedup, the `(method, url)` key in `optimize_endpoints` could be first seen from either analyzer depending on fiber completion order. The `flask_foreign` test's assertion on `.details.technology == "python_fastapi"` therefore flaked (seen as `python_flask` in https://github.com/owasp-noir/noir/actions/runs/26984272313/job/79630377888).

- Comment-strip (`gsub(/#.*?(?:\n|\z)/m, "\n")`) before all relevance heuristics so "flask" in comments/docs no longer triggers false positive.
- The competing import guard now reliably fires for pure FastAPI/Sanic/etc. files even when the Flask detector fired on a sibling file in the tree.
- Decorator and `add_url_rule` fallbacks also now ignore commented-out code.
- 20+ repeated runs of the specific spec + full suite + manual `./bin/noir` run on the fixture all confirm stable `python_fastapi` attribution with no cross-claim duplicates.

This makes the foreign-framework guard (and the per-file relevance filters used by other Python analyzers) robust.

Fixes the test breakage reported in the linked CI run.

### DIFF
--- a/src/analyzer/analyzers/python/flask.cr
+++ b/src/analyzer/analyzers/python/flask.cr
@@ -790,7 +790,15 @@ module Analyzer::Python
     end
 
     private def flask_relevant_source?(source : ::String) : Bool
-      return true if source.includes?("flask")
+      # Strip # comments (to EOL) before relevance heuristics. This prevents
+      # explanatory comments (in fixtures, docs, or real code) that mention
+      # framework names from causing mis-attribution. E.g. a FastAPI file
+      # whose comment says "flask" would otherwise bypass the competing-import
+      # guard and emit duplicate routes under python_flask tech (which then
+      # wins dedup non-deterministically depending on fiber completion order).
+      clean = source.gsub(/#.*?(?:\n|\z)/m, "\n")
+
+      return true if clean.includes?("flask")
       # A file that imports a competing decorator-based framework (and
       # never mentions flask) is NOT Flask — its `@app.get`/`@router.post`
       # decorators belong to that framework's analyzer. Without this
@@ -801,9 +809,9 @@ module Analyzer::Python
       # params. The route's real owner still emits it correctly, so this
       # only drops the duplicate mislabel (jupyterhub examples/service-
       # fastapi was reported as python_flask).
-      return false if source.matches?(/^\s*(?:from|import)\s+(?:fastapi|sanic|litestar|starlette|quart|robyn)\b/m)
-      return true if source.matches?(/^\s*@\s*#{DOT_NATION}\s*\.\s*(?:route|get|post|put|patch|delete|head|options|trace)\s*\(/m)
-      return true if source.matches?(/\b#{DOT_NATION}\s*\.\s*(?:add_url_rule|register_blueprint)\s*\(/)
+      return false if clean.matches?(/^\s*(?:from|import)\s+(?:fastapi|sanic|litestar|starlette|quart|robyn)\b/m)
+      return true if clean.matches?(/^\s*@\s*#{DOT_NATION}\s*\.\s*(?:route|get|post|put|patch|delete|head|options|trace)\s*\(/m)
+      return true if clean.matches?(/\b#{DOT_NATION}\s*\.\s*(?:add_url_rule|register_blueprint)\s*\(/)
 
       false
     end


src/analyzer/analyzers/python/flask.cr
```diff
@@ -790,7 +790,15 @@ module Analyzer::Python
...
-      return true if source.includes?("flask")
+      # Strip # comments (to EOL) before relevance heuristics. This prevents
+      # explanatory comments (in fixtures, docs, or real code) that mention
+      # framework names from causing mis-attribution. E.g. a FastAPI file
+      # whose comment says "flask" would otherwise bypass the competing-import
+      # guard and emit duplicate routes under python_flask tech (which then
+      # wins dedup non-deterministically depending on fiber completion order).
+      clean = source.gsub(/#.*?(?:\n|\z)/m, "\n")
+
+      return true if clean.includes?("flask")
```
hahwul, line 801: **[bug]** `return true if clean.includes?("flask")` executes before the competing import guard. Because only `#...` comments are stripped into `clean`, a "flask" (lowercase) substring inside a docstring or any string literal on a foreign-framework file (e.g. `"""This is a flask-like example using FastAPI"""` or `msg = "see flask docs"`) will still cause an early true, making the Flask analyzer process the file and emit duplicate routes (bypassing `return false if clean.matches?(... fastapi ...)`). The foreign-framework claim bug is therefore only partially fixed.

**Suggestion:** Evaluate the competing-import `return false` (or a more general "this file imports a known non-Flask decorator framework") before or regardless of any "flask" mention; or strengthen the "flask" test to ignore occurrences inside quotes and docstrings (for the heuristic only); or document that only `#` comments are mitigated.

src/analyzer/analyzers/python/flask.cr
```diff
@@ -790,7 +790,15 @@ module Analyzer::Python
...
 
     private def flask_relevant_source?(source : ::String) : Bool
-      return true if source.includes?("flask")
+      # Strip # comments (to EOL) before relevance heuristics. This prevents
+      # explanatory comments (in fixtures, docs, or real code) that mention
+      # framework names from causing mis-attribution. E.g. a FastAPI file
+      # whose comment says "flask" would otherwise bypass the competing-import
+      # guard and emit duplicate routes under python_flask tech (which then
+      # wins dedup non-deterministically depending on fiber completion order).
+      clean = source.gsub(/#.*?(?:\n|\z)/m, "\n")
```
hahwul, line 799: **[suggestion]** `clean = source.gsub(/#.*?(?:\n|\z)/m, "\n")` unconditionally allocates a fresh String for every `.py` file the Flask analyzer walks (even when `source` contains no `#`). While negligible for typical projects, it adds work on the hot path before the subsequent cheap `includes?`/`matches?` checks. The rest of the file already uses per-line `lstrip.starts_with?("#")` skips inside the `if flask_relevant_source?` block.

**Suggestion:** Add a fast-path `return true if source.includes?("flask") unless source.includes?("#")` (or equivalent) or perform the strip only for the three checks that need a comment-free view; consider sharing a comment-stripped view with the line-iteration logic already present.

src/analyzer/analyzers/python/flask.cr
```diff
@@ -790,7 +790,15 @@ module Analyzer::Python
     end
 
     private def flask_relevant_source?(source : ::String) : Bool
-      return true if source.includes?("flask")
+      # Strip # comments (to EOL) before relevance heuristics. This prevents
```
hahwul, line 793: **[nit]** The added comment claims the change prevents "explanatory comments (in fixtures, docs, or real code) that mention framework names". In reality only `#`-to-EOL constructs are removed; "flask" text inside `"""..."""`, `'''...'''`, `"..."`/`'...'` (docstrings, data strings, "ported from flask" copy-pasta, etc.) on a foreign file will still short-circuit to true and bypass the competing-import guard.

**Suggestion:** Edit the comment to say "Strip Python `#` comments (to EOL) ..." and qualify the protection (e.g. "hash comments in fixtures...") so readers are not misled about the scope.

src/analyzer/analyzers/python/flask.cr
```diff
@@ -790,7 +790,15 @@ module Analyzer::Python
     end
 
     private def flask_relevant_source?(source : ::String) : Bool
```
hahwul, line 792: **[suggestion]** `flask_relevant_source?` is private, has no direct unit tests (only exercised indirectly by `spec/functional_test/testers/python/flask_foreign_spec.cr` plus other Flask fixtures), and the new comment-stripping logic plus its edge cases (docstring mentions, CRLF, inline `code # flask`, EOF comment, "flask" inside a string arg to a decorator, etc.) have no dedicated coverage.

**Suggestion:** Add a small unit test (under `spec/unit_test/analyzer/` or by extending an existing Flask analyzer test) that calls the method (via `send` or by making it protected for testing) with the cases above to prevent future regressions in the heuristic.

hahwul (commented):
## Summary

The PR narrows the Flask Python analyzer's `flask_relevant_source?` gate (the only change; `src/analyzer/analyzers/python/flask.cr`) by pre-stripping `#`-to-EOL comments before the `includes?("flask")`, competing-decorator-framework import, and `@<var>.(route|get|...)` / `add_url_rule` heuristics. The goal is to stop explanatory comments that name "flask" (as in the accompanying `flask_foreign/service.py` fixture and its spec) from causing the Flask analyzer to claim routes that belong to FastAPI/Sanic/etc. when a project mixes frameworks. The functional test now passes, the motivation comment is clear, and the regex change is correctly non-greedy.

Correctness is mostly sound for the `#` comment case that was reported, but the early `includes?("flask")` return still precedes the competing-import guard and only `#` comments are removed, leaving string/docstring mentions as an unaddressed vector. Similar raw-name checks exist in peer Python analyzers but are out of scope for this diff.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 1

PR #1904, opened 2026-06-05
Fix GraphQL SDL detection for operation documents

### LINKED ISSUE #1902: GraphQL SDL analyzer consumes excessive CPU on large GraphQL operation documents
### Summary

Noir detects some `.graphql` operation documents as `graphql_sdl` and then spends a very long time in the SDL analyzer without producing useful endpoints.

### Reproduction

In a Next.js monorepo with generated GraphQL operation documents under `packages/bff/src/**`, run:

```sh
noir scan . --only-techs graphql_sdl --no-spinner --no-color --no-log
```

A representative file is `packages/bff/src/swg2gql/partner/partner.graphql`:

- about 329 KB
- about 14,053 lines
- contains many top-level `query ... { ... }` and `mutation ... { ... }` operation documents
- is not an SDL schema document with `type Query { ... }`

Scanning only the directory that contains that file was still running after 30 seconds and had to be interrupted. Scanning the project with `graphql_sdl` enabled also spent over a minute of CPU time in this path before I stopped it.

### Observed behavior

`graphql_sdl` is enabled because the detector finds SDL-like tokens in `.graphql` files. The analyzer then parses generated operation documents as if they were SDL. On large generated operation documents this becomes a major CPU sink and produces no useful SDL endpoint output.

### Expected behavior

Noir should quickly skip GraphQL operation documents, or the SDL analyzer should avoid expensive work when a `.graphql` file is not an SDL schema. Operation documents can be handled by a separate analyzer if desired, but they should not make the SDL analyzer slow.

### Notes

A practical mitigation today is to disable `graphql_sdl` for this repository:

```sh
noir scan . --exclude-techs graphql_sdl
```

한국어 요약: 큰 GraphQL operation 문서가 SDL로 감지되어 `graphql_sdl` 분석이 과도하게 오래 걸립니다.

hahwul:
### Summary

- tighten GraphQL SDL detection so SDL keywords are only accepted as top-level declarations
- avoid treating GraphQL operation selection fields such as `type`, `schema`, or `enum` as SDL schema signals
- add a regression test for operation documents containing SDL keyword field names

Fixes #1902.

### Verification

- `crystal spec spec/unit_test/detector/specification/graphql_sdl_spec.cr`
- `crystal spec spec/unit_test/detector/specification/graphql_sdl_spec.cr spec/unit_test/analyzer/specification/graphql_sdl_spec.cr spec/functional_test/testers/specification/graphql_sdl_spec.cr`
- `just check`
- `just build`
- `./bin/noir scan /Users/pluto.h/z_app/test/guest-front/packages/bff/src/swg2gql/partner --only-techs graphql_sdl --no-spinner --no-color --no-log` completed in ~0.95s

### DIFF
--- a/spec/unit_test/detector/specification/graphql_sdl_spec.cr
+++ b/spec/unit_test/detector/specification/graphql_sdl_spec.cr
@@ -16,6 +16,24 @@ describe "Detect GraphQL SDL" do
     instance.detect("schema.graphql", content).should be_true
   end
 
+  it "detects SDL with regular-string description on same line as declaration" do
+    content = <<-SDL
+      "Description for the root" type Query {
+        hello: String
+      }
+      SDL
+
+    instance.detect("schema.graphql", content).should be_true
+  end
+
+  it "detects SDL with block-string description immediately followed by declaration on same line" do
+    content = <<-SDL
+      """Block desc""" type Query { hello: String }
+      SDL
+
+    instance.detect("schema.graphql", content).should be_true
+  end
+
   it ".graphqls with schema block" do
     content = <<-SDL
       schema {
@@ -51,6 +69,24 @@ describe "Detect GraphQL SDL" do
   it "rejects operation documents with SDL keyword field names" do
     content = <<-GQL
       query GetStore {
+        store {
+          id
+          type
+          schema {
+            enum
+          }
+        }
+      }
+      GQL
+
+    instance.detect("ops.graphql", content).should be_false
+  end
+
+  it "rejects operation documents even when string literals contain braces, #, or SDL keywords" do
+    # Strings can contain example SDL, # (not comments), or unbalanced braces.
+    # These must not affect depth tracking or cause false SDL detection.
+    content = <<-GQL
+      query GetStore($filter: Filter = "{ schema: 1 } # not a comment", $desc: String = "type Query { x }") {
         store {
           id
           type
--- a/spec/unit_test/detector/specification/graphql_sdl_spec.cr
+++ b/spec/unit_test/detector/specification/graphql_sdl_spec.cr
@@ -48,6 +48,22 @@ describe "Detect GraphQL SDL" do
     instance.detect("ops.graphql", content).should be_false
   end
 
+  it "rejects operation documents with SDL keyword field names" do
+    content = <<-GQL
+      query GetStore {
+        store {
+          id
+          type
+          schema {
+            enum
+          }
+        }
+      }
+      GQL
+
+    instance.detect("ops.graphql", content).should be_false
+  end
+
   it "rejects non-graphql files" do
     instance.applicable?("schema.json").should be_false
   end
--- a/src/detector/detectors/specification/graphql_sdl.cr
+++ b/src/detector/detectors/specification/graphql_sdl.cr
@@ -3,17 +3,66 @@ require "../../../models/code_locator"
 
 module Detector::Specification
   class GraphqlSdl < Detector
-    # Top-level SDL signals — distinguishes a schema document from an
-    # operation document (which the file_analyzers/graphql_analyzer handles).
-    SDL_SIGNAL = /^[\s]*(?:extend\s+)?(?:type|input|interface|union|enum|scalar|directive|schema)\b/m
+    # SDL declaration signals — distinguished at top-level only by
+    # `sdl_document?`. A broad keyword-only check is too loose: generated
+    # operation documents often select fields named `type`, `schema`, or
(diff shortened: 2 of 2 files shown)


hahwul:
@claude @copilot+claude-opus-4.6 
Let's review the code.

hahwul (commented):
## Review — looks good ✅ (posting as COMMENT; GitHub blocks self‑approval)

Fixing this in the **detector** is the right layer. Issue #1902 is really a performance problem (the SDL *analyzer* burning CPU on a 329 KB / 14 k‑line operation document). By making `detect` return `false` for operation documents, the file is never pushed into `CodeLocator("graphql-sdl")`, so `Analyzer::Specification::GraphqlSdl` never runs the expensive `GraphqlSdlParser` on it. Root cause addressed, not just the symptom.

### What I checked
- **No SDL keyword lost vs. the old `SDL_SIGNAL`.** Old set `type|input|interface|union|enum|scalar|directive|schema` is all still reachable: `type/input/interface/enum/scalar` now require a following identifier, `union` requires `… =`, `directive` requires `@name`, `schema` stays bare. Positive specs (`type Query`, `schema {…}`, `extend type`) still pass.
- **Block‑string description above a type still detects** — traced `"""\n…\n"""\ntype Query {` → `in_block_string` toggles correctly and `type Query {` matches at depth 0. 👍
- **The regression test exercises both guards**, which is nice: bare `type`/`enum` are rejected by the stricter regex, but `schema {` *does* match `schema\b` — it's only rejected because it sits at `depth == 2`. So the test genuinely validates the depth tracking, not just the regex.
- CI `tests (1.19.0)` / `tests (1.20.0)` green.

### Non‑blocking notes (optional)
1. **Depth tracking only special‑cases block strings (`"""`), not ordinary `"…"` strings or `#` inside strings.** `update_depth` counts every `{`/`}`, and `detection_line` treats the first `#` as a comment. So a string literal containing a stray brace (e.g. a default value `arg: "}"`) or a `#` can mis‑track depth / truncate a line. For a real *false positive* you'd need the running depth to land back on 0 **and** a line matching `keyword identifier` (or `schema {`) — constructible but pathological, and generated operation docs like the one in #1902 don't do this. Since the regex is the primary guard, impact is low. Just flagging it as a known limitation.
2. **Multiple `"""` on a single line** (open + close + reopen) isn't fully modeled in `detection_line` — it returns the text before the first `"""` and marks the string closed, which can drop trailing text. Extremely rare in practice.
3. **Perf worst case shifts to a full scan.** For a large operation document (no SDL), the detector can't early‑out and now walks every line + every char to return `false`. That's bounded and cheap (your ~0.95 s end‑to‑end confirms it) and far cheaper than the old behavior of handing the file to the SDL parser — net win. Noting it only so it's a conscious trade‑off.

None of these block merge. Nice, well‑tested fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

hahwul:
**Improvements based on Copilot + Claude review feedback**

Addressed the non-blocking notes from the Claude review (the one posted as COMMENT on the original commit):

1. **Regular strings and `#` inside strings** — `detection_line` now uses a char-by-char state machine that tracks *both* `"""` block strings **and** regular `"..."` strings (with basic `\` escape support). `#` is only treated as comment when not inside a string.
2. **Text after closing block strings on the same line** — now collected for both candidate matching and depth (e.g. `"""desc""" type Query { ... }` or `"""desc""" schema { ... }` will correctly see the declaration).
3. Depth tracking now only counts `{`/`}` that appear in the *visible code* (strings excised), so default values like `arg: "{ foo: \"}\" }"` or strings with example SDL no longer corrupt the depth counter.

**Added test coverage** (in `graphql_sdl_spec.cr`):
- Detector now explicitly tests regular-string and block-string descriptions preceding declarations (including same-line cases).
- New regression case for operation documents whose *operation signature* contains string literals with `#`, unbalanced `{}`, and SDL keywords inside them (e.g. complex default values). These previously could have caused false-positive SDL detection due to early truncation or depth underflow.

**Verification performed locally** (before push):
- `crystal spec spec/unit_test/detector/specification/graphql_sdl_spec.cr` (now 10 examples)
- `crystal spec .../graphql_sdl_spec.cr` (analyzer unit + functional)
- `just check` (format + ameba) — clean
- `just build` — clean
- Full `just test` — 30k+ examples, 0 failures

The original fix (stricter decl regex + depth==0 guard) is unchanged in spirit; this just makes the supporting lexer more correct per the review.

CI is currently running on the update (new commit). Once green, we can merge.

(Also pushed the improvement commit to the branch.)

PR #1910, opened 2026-06-05
fix(tagger): cut FP/FN in pii, soap, webhook, websocket taggers

hahwul:
## Summary

Tightens four metadata taggers in `src/tagger/taggers` to enrich Noir's AI/review context with fewer misses and fewer false flags. These taggers feed downstream consumers that prioritize sensitive routes, so both recall (FN) and precision (FP) matter. Each change is grounded in how the analyzers actually emit endpoints (param casing, header `param_type`, protocol metadata).

### PII (`pii.cr`)
- **FN — camelCase normalization.** `normalize_param_name` only did `downcase` + `-`→`_`, so analyzers that emit `firstName`, `dateOfBirth`, `emailAddress`, `postalCode`, … (JS/Java/etc.) were **silently missed** because the lists are snake_case. Now camelCase boundaries fold to snake_case first (`firstName` → `first_name`), with digit runs left intact so `cvv2` survives.
- **FN — strong tokens in compound names.** A new `STRONG_TOKENS` set matches unambiguous single tokens (`ssn`, `cvv`, `iban`, `passport`, `dob`, …) anywhere in a normalized name, catching `customerSsn`, `applicantPassport`, etc. Token-level (not substring) matching keeps `lessonId` from tripping on the `ssn` substring.
- **FN — list coverage.** Added card/passport/tax/national-id/address/phone/name variants.

### SOAP (`soap.cr`)
- **FN — URL + content-type.** Only the `SOAPAction` header was detected. Now also flags `?wsdl` / `.wsdl` / `.asmx` URLs and the SOAP 1.2 `application/soap+xml` `Content-Type`.
- **FP guard.** `SOAPAction` is matched only as a `header` param; a bare `soap` path segment stays excluded (e.g. `/products/soap` store listing).

### Webhook (`webhook.cr`)
- **FP — drop `notification(s)` path parts.** An in-app notifications resource (`POST /api/notifications`, `DELETE /notifications/{id}`) is a write, so the existing not-a-read gate couldn't save it from being mistagged as a webhook. The verb `notify` is kept (canonical payment-IPN `notify_url`).
- **FN — broader signature headers.** Added Shopify, Svix (Clerk/Resend), Razorpay, Paystack, Square, Twilio, Line, Zoom, Discord (`X-Signature-Ed25519`), Telegram, WooCommerce, SNS, Mandrill, and generic `webhook_signature`/`x_hook_signature`.

### WebSocket (`websocket.cr`)
- **FN — conclusive single header.** `Sec-WebSocket-Key`/`-Accept` are reserved for the handshake, so one is now enough (previously **2** of any sec-websocket-* were required). An explicit `Upgrade: websocket` also flags. Weaker handshake headers (version/protocol/extensions) still need two.
- **FN — transport markers.** Socket.IO / SockJS run an HTTP handshake before upgrading, so their routes arrive as plain-HTTP endpoints; `socket.io` / `sockjs` URLs are now detected.

## Test plan
- Added/updated specs for every change above (camelCase, compound-token + substring-FP guard, WSDL/asmx/soap+xml, soap-segment FP, notifications FP, new signature headers, single-strong-header, Upgrade, Socket.IO/SockJS).
- `crystal spec spec/unit_test/tagger/` → 427 examples, 0 failures.
- `crystal spec spec/unit_test/` → 3035 examples, 0 failures.
- Full `crystal build src/noir.cr` compiles; ameba clean on all changed files.

### DIFF
--- a/spec/unit_test/tagger/taggers/crypto_spec.cr
+++ b/spec/unit_test/tagger/taggers/crypto_spec.cr
@@ -83,7 +83,7 @@ describe "CryptoTagger" do
 
       tagger.perform(endpoints)
 
-      endpoints.each { |ep| ep.tags.map(&.name).should contain("crypto") }
+      endpoints.each(&.tags.map(&.name).should(contain("crypto")))
     end
 
     it "tags a JOSE jwe/jws path" do
--- a/spec/unit_test/tagger/taggers/pii_spec.cr
+++ b/spec/unit_test/tagger/taggers/pii_spec.cr
@@ -100,5 +100,60 @@ describe "PiiTagger" do
       endpoint.tags.size.should eq(1)
       endpoint.tags[0].name.should eq("pii")
     end
+
+    it "normalizes camelCase strong names (dateOfBirth)" do
+      tagger = PiiTagger.new(default_tagger_options)
+
+      endpoint = Endpoint.new("/api/kyc", "POST", [
+        Param.new("dateOfBirth", "", "json"),
+      ])
+
+      tagger.perform([endpoint])
+
+      endpoint.tags.size.should eq(1)
+      endpoint.tags[0].name.should eq("pii")
+    end
+
+    it "tags two camelCase medium names (firstName + lastName)" do
+      tagger = PiiTagger.new(default_tagger_options)
+
+      endpoint = Endpoint.new("/api/profile", "POST", [
+        Param.new("firstName", "", "json"),
+        Param.new("lastName", "", "json"),
+      ])
+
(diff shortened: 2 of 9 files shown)


hahwul:
Rebased onto latest main and fixed an unrelated lint regression that landed in #1906: `Style/VerboseBlock` in `spec/unit_test/tagger/taggers/crypto_spec.cr:86` (was failing CI lint for every PR off current main). Full-repo `ameba` now reports 1338 inspected, 0 failures.

PR #1933, opened 2026-06-06
perf(analyzer): improve Kotlin Spring extraction

hahwul:
## Summary
- skip Kotlin DTO index construction for files that do not produce Spring routes
- reuse per-file local string constants and pre-index handler method nodes to avoid repeated AST walks per route
- recover routes/params from Kotlin controllers that use split constructor annotations (`class C` followed by `@Autowired constructor(...)`), while keeping abstract split-constructor base controllers ignored

## Validation
- `just check`
- `just test`
- scanned Kotlin Spring OSS samples: `gs-testing-web`, `kotlin-spring-boot-mvc-starter`, `rest-controller-kotlin`, `spring-04-controller`, `spring-kotlin-web-crud`, `spring-simplicity-kotlin`; normalized endpoint output remained unchanged
- verified a split-constructor controller now emits `GET /api/{id}` with `q` query and `id` path params

### DIFF
--- a/spec/unit_test/miniparser/kotlin_parameter_extractor_ts_spec.cr
+++ b/spec/unit_test/miniparser/kotlin_parameter_extractor_ts_spec.cr
@@ -236,6 +236,31 @@ describe Noir::TreeSitterKotlinParameterExtractor do
         {"userIds", "query"},
       ])
     end
+
+    it "indexes methods recovered from split constructor annotations" do
+      source = <<-KT
+        package com.example
+
+        @RestController
+        class C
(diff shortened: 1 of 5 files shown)


hahwul (commented):
LGTM.

Perf improvement for Kotlin Spring analyzer:
- Skips DTO indexing for non-route files, reuses per-file indexes, recovers split-constructor @Autowired cases.
- Added fixtures/tests implied by PR.
- Format check passed in worktree.
- Matches miniparser layering (kotlin_*_extractor_ts.cr updates).
CI tests passed. Good optimization work.

PR #1968, opened 2026-06-10
feat(mobile): iOS Info.plist / entitlements analyzer (#1943)

hahwul:
## Summary

Second analyzer in the mobile series for **#1943**. Adds iOS deep-link extraction, using the **same model** established for Android (one endpoint per URI, `method = "GET"`, mobile semantics in `protocol`, `metadata` suppressed when nil).

> **Stacked on #1967** (Android). Base branch is `mobile-android-analyzer`; please merge #1967 first (GitHub will retarget this to `main` automatically). The diff below is iOS-only.

## What it parses

| Source | protocol | url |
|---|---|---|
| `Info.plist` `CFBundleURLTypes` → `CFBundleURLSchemes` | `mobile-scheme` | `myapp://` |
| `*.entitlements` `associated-domains` `applinks:<domain>` | `universal-link` | `https://<domain>/` |

- iOS schemes declare no path, so the URL is scheme-only (`myapp://`).
- `applinks:example.com?mode=developer` → `https://example.com/` (mode suffix stripped); `webcredentials:` / `activitycontinuation:` entries are ignored.
- XML plists are parsed with the built-in `XML` module; binary plists fail gracefully (debug log, no crash).
- The detector gates on `CFBundleURLTypes` / `associated-domains` content so ordinary plists don't register a phantom `ios` tech.

## Handler linkage (note)

iOS config files name no handler component (deep links are dispatched by the App/SceneDelegate `onOpenURL` / `application(_:open:)`), so `metadata["via"]` is intentionally empty here — it's left for the upcoming **code-linkage layer**, which will resolve the handler, pull intent/URI params, and populate `callees` + `ai_context` (the taint path from the incoming URL to its sinks). Same intent as Android's `via`.

## Tests

- Fixture `spec/functional_test/fixtures/mobile/ios/` (`Info.plist` + `App.entitlements`) covering multi-scheme URL types, applinks, the `?mode=` strip, and the webcredentials-ignored case; tester asserts 4 endpoints + protocols (1 tech).
- Unit spec `spec/unit_test/analyzer/mobile_ios_spec.cr` for the plist-walking edge cases.
- Full suite green (`20990 examples, 0 failures`); `ameba` + `crystal tool format` clean.

## Next
[...]

Issue #1943, opened 2026-06-08
Analyzer for Android/iOS App


hahwul:
@ksg97031 Any ideas?

## Result Example

```
SCHEME myapp://complex/:id?SearchQuery=
  ○ type: mobile-scheme
  ○ path: id
  ○ query: SearchQuery
  ○ intent: android.intent.action.VIEW
  ○ host: complex
  ○ package: com.example.myapp   # Optional

INTENT com.example.myapp/.DeepLinkActivity
  ○ type: android-intent
  ○ action: android.intent.action.VIEW
  ○ data: myapp://accounts/:id
  ○ category: android.intent.category.BROWSABLE
  ○ path: id
  ○ extras: {"userId": ""}       # Intent extras

UNIVERSAL https://myapp.example.com/complex/:id
  ○ type: universal-link
  ○ path: id
```

PR #1969, opened 2026-06-10
feat(mobile): link Android deep links to handler code — callees, input params, ai_context (#1943)

hahwul:
## Summary

The core of **#1943**: a deep-link endpoint should not be a dead config-derived entry — it should connect to the code that handles it, so `callees` and `ai_context` (the taint path) light up. This PR wires Android deep links to their handler source.

> **Stacked on #1967** (Android analyzer). Base is `mobile-android-analyzer`; merge #1967 first. iOS code linkage is a follow-up (iOS config names no handler component, so it needs `onOpenURL`/SceneDelegate discovery).

## How it works

A new post-analysis pass `NoirMobileLinker.apply(@endpoints)` runs after the optimizer, before taggers + AI-context. For each Android mobile endpoint it:

1. **Resolves the handler component** — `metadata["via"]` (scheme / universal-link) or the `intent://<pkg>/<component>` URL (android-intent) → the declaring `.kt`/`.java` file, via a lazy class→file index.
2. **Adds the handler file as a `code_path`** so the AI-context builder scans the handler body.
3. **Extracts 1-hop callees** from the intent-handling methods (`onCreate`, `onNewIntent`, `onReceive`, `onStartCommand`, …) using the existing tree-sitter Kotlin/Java callee extractors.

The existing `AIContext` builder (which is **not** tech-gated — it keys on `code_paths` + `callees`) then derives sinks/guards/sources for free. Two mobile-specific sink patterns are added to `patterns.cr`:
- `webview_load` — `WebView.loadUrl` / `loadData` / `evaluateJavascript` (deep-link XSS/SSRF surface)
- `intent_redirect` — `startActivity` / `sendBroadcast` from inbound data

## Result

For the fixture's `myapp://complex/:id` (handled by `DeepLinkActivity.onCreate`):

```
callees:  super.onCreate, data.getQueryParameter, renderProfile, webView.loadUrl
code_paths: AndroidManifest.xml, DeepLinkActivity.kt
ai_context: sinks=[webview_load], callees=4, sources=[request_input]
```

Endpoints whose component has no resolvable source stay config-only (no false enrichment).

## Notes / fixes

- **Struct `.tap` footgun:** `Endpoint.new(...).tap(&.protocol = "...")` does **not** persist (the block mutates a copy), so the mobile functional testers were silently *skipping* the `protocol`/`callee` assertions (FunctionalTester guards on `expected.protocol != "http"`). Rebuilt expected endpoints via a Proc that returns the mutated struct, so those assertions now actually run. (The iOS tester on #1968 has the same latent issue — I'll fix it on that branch.)
- `read_content` falls back to `File.read` when the content cache is cold (budget exhausted / disabled), per the `CodeLocator` contract.
- `android` added to `CALLEE_SUPPORTED_TECHS` / `AI_CONTEXT_GUARD_SUPPORTED_TECHS` (display-only lists).

## Tests
- Fixture gains `src/main/java/com/example/myapp/DeepLinkActivity.kt`; the android tester asserts the extracted callees; unit spec `spec/unit_test/mobile/linker_spec.cr` covers class resolution, callee extraction, code_path attach, and the no-source no-op.
- Full suite green (`20985 examples, 0 failures`); `ameba` + format clean.

## Next
- iOS code linkage (`onOpenURL` / `application(_:open:)`).
- Fast-follow: extract intent extras / `getQueryParameter` args as `Param`s (explicit input surface + stronger source signal).
- Guards/hardening PR (curl/OAS/deliver skip), docs prose page.

Part of #1943.

### DIFF
--- a/spec/functional_test/fixtures/mobile/android/src/main/java/com/example/myapp/DeepLinkActivity.kt
+++ b/spec/functional_test/fixtures/mobile/android/src/main/java/com/example/myapp/DeepLinkActivity.kt
@@ -1,0 +1,24 @@
+package com.example.myapp
+
+import android.app.Activity
+import android.os.Bundle
+import android.webkit.WebView
+
+class DeepLinkActivity : Activity() {
+    private lateinit var webView: WebView
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val data = intent.data
+        val target = data?.getQueryParameter("redirect")
+        val id = data?.lastPathSegment
+        renderProfile(id)
+        if (target != null) {
+            webView.loadUrl(target)
+        }
+    }
+
+    private fun renderProfile(id: String?) {
+        webView.loadData("<h1>" + id + "</h1>", "text/html", "utf-8")
+    }
+}
--- a/spec/functional_test/fixtures/mobile/android/src/main/java/com/example/myapp/DeepLinkActivity.kt
+++ b/spec/functional_test/fixtures/mobile/android/src/main/java/com/example/myapp/DeepLinkActivity.kt
@@ -11,14 +11,17 @@ class DeepLinkActivity : Activity() {
         super.onCreate(savedInstanceState)
         val data = intent.data
         val target = data?.getQueryParameter("redirect")
+        val campaign = data?.getQueryParameter("utm_source")
+        val userId = intent.getStringExtra("userId")
+        val verified = intent.getBooleanExtra("verified", false)
         val id = data?.lastPathSegment
-        renderProfile(id)
+        renderProfile(id, userId)
         if (target != null) {
             webView.loadUrl(target)
         }
     }
 
-    private fun renderProfile(id: String?) {
+    private fun renderProfile(id: String?, userId: String?) {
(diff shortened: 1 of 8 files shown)


hahwul:
Pushed a follow-up: the linker now also extracts the **handler's input params** from the same parse.

- `uri.getQueryParameter("redirect")` → `query` param (bakes into the URL, e.g. `myapp://complex/:id?redirect=&utm_source=`, matching the issue mock).
- `intent.getStringExtra("userId")` / `getBooleanExtra(...)` etc. → `extra` param (Intent extras are a Bundle, not part of the URI); rendered as `○ extras: userId, verified`.

Full picture now on `myapp://complex/:id`:
```
params:  id(path), redirect(query), utm_source(query), userId(extra), verified(extra)
sources (ai_context): all 5 → request_input
sinks (ai_context):   webview_load
callees:  webView.loadUrl, renderProfile, data.getQueryParameter, ...
```
So the deep link's full untrusted-input surface is wired to its sink. Spec count 20995, green.

hahwul:
Superseded by #1971 (re-created on `main` after the stack's base branch was deleted). Same content + review fixes; merged there.

PR #2195, opened 2026-06-21
Use idiomatic `unless x.nil?` instead of `if !x.nil?`

### LINKED ISSUE #2169: Use idiomatic `unless x.nil?` instead of `if !x.nil?` across the analyzers
## Description

Around 26 clean single-condition sites (heavily concentrated in the Python analyzers, plus a few output builders, models, and miniparsers) write `if !x.nil?` where the idiomatic Crystal form is `unless x.nil?`. The codebase already prefers `unless x.nil?` (12 existing sites) and `if x.nil?` (465 sites), so this is a true consistency cleanup.

### Affected files (clean single-condition sites)

- `src/analyzer/analyzers/python/django.cr` — :1240, :1242, :1252, :1274, :1332, :1334, :1820, :1859 (+others)
- `src/analyzer/analyzers/python/flask.cr` — :1470, :1533
- `src/analyzer/analyzers/python/fastapi.cr` — :76, :103, :192
- `src/analyzer/analyzers/python/quart.cr:819`, `src/analyzer/analyzers/python/sanic.cr:963`
- `src/analyzer/engines/python_engine.cr:183`
- `src/output_builder/only-tag.cr:8`, `src/output_builder/markdown_table.cr:10`, `src/models/output_builder.cr:85`, `src/models/analyzer.cr:197`
- `src/miniparsers/import_graph.cr:351, :459`
- `src/analyzer/analyzers/javascript/nestjs.cr:356`

### Proposed change

```crystal
# before
if !match.nil?
  do_something(match[1])
end

# after
unless match.nil?
  do_something(match[1])
end
```

Prefer `unless x.nil?` over `if x` — for a value whose type includes `false`, `if x` would change truthiness, whereas `unless x.nil?` preserves exact nil-narrowing semantics so bodies like `match[1]` keep compiling. Scope to the simple block-`if` sites (skip compound `&&`/`||` and `elsif` cases).

### Why this is a good first issue

A mechanical, site-by-site syntactic swap the compiler verifies (nil-narrowing is unchanged), with the target idiom already present in the codebase to copy from. No detection logic, regexes, or route/param extraction is touched.

---
🇰🇷 분석기 전반의 `if !x.nil?`를 관용적인 `unless x.nil?` 형태로 바꾸는 일관성 정리 작업입니다.

gords2:
## Description

Closes #2169.

Replaces the negated-nil `if !x.nil?` form with the idiomatic Crystal `unless x.nil?` at **simple single-condition block-`if` sites** across the analyzers, output builders, models, and miniparsers. The codebase already prefers `unless x.nil?`, so this is a consistency cleanup with identical nil-narrowing semantics.

**Scope discipline:** only clean single-condition sites were converted (19 in total). Intentionally left as `if !x.nil?`:
- compound conditions (`&& / ||`)
- `elsif` branches
- inline-modifier `… if !x.nil?`
- any block that carries an `else`/`elsif` — converting those would produce an `unless … else`, which is non-idiomatic and flagged by Ameba's `Style/UnlessElse`.

(7 sites that initially matched the simple pattern but turned out to have an `else`/`elsif` were caught by Ameba and reverted.)

## Verification

- `crystal tool format` — clean
- `shards build` — passes
- `crystal run lib/ameba/bin/ameba.cr` on all changed files — `0 failures`
- `crystal spec spec/unit_test` — `3615 examples, 0 failures`
- `crystal spec spec/functional_test` — `19809 examples, 0 failures`

### DIFF
--- a/src/analyzer/analyzers/javascript/nestjs.cr
+++ b/src/analyzer/analyzers/javascript/nestjs.cr
@@ -353,7 +353,7 @@ module Analyzer::Javascript
         if line.includes?("@Controller")
           joined = join_decorator_header(lines, index)
           base_paths = parse_controller_decorator(joined[:text], literal_values)
-          if !base_paths.nil?
+          unless base_paths.nil?
             current_base_paths = base_paths
             current_versions = parse_controller_versions(joined[:text])
             current_content.clear
--- a/src/analyzer/analyzers/python/django.cr
+++ b/src/analyzer/analyzers/python/django.cr
@@ -1236,9 +1236,9 @@ module Analyzer::Python
 
       # Function Based View
       function_start_index = content.index /(?:async\s+)?def\s+#{function_or_class_name}\s*\(/
-      if !function_start_index.nil?
+      unless function_start_index.nil?
         function_codeblock = parse_code_block(content[function_start_index..])
-        if !function_codeblock.nil?
+        unless function_codeblock.nil?
           lines = function_codeblock.split "\n"
           function_define_line = lines[0]
           lines = lines[1..]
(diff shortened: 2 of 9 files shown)


hahwul (approved):
LGTM :)

PR #2202, opened 2026-06-22
test: add direct contract tests for join_path utility

### LINKED ISSUE #2191: Add a spec for the global `join_path` path-join helper in `src/utils/utils.cr`
## Description

`src/utils/utils.cr` is a tidy, well-tested little utility module — almost every helper in it already has a dedicated unit test. One pure helper, the global variadic `join_path`, is the lone exception and currently has no spec. This is a clean, behavior-preserving polish: add a `describe "join_path"` block that pins down its existing, verified behavior. No production code changes.

`join_path` is a pure `String... -> String` function (no IO, no ENV, no framework state) that composes URL path segments — it strips per-segment slashes and guarantees a single leading slash. It is used by the Clojure analyzers (reitit/compojure/pedestal) to build route prefixes, so locking in its behavior with a spec is worthwhile.

### Affected files
- `src/utils/utils.cr:34` — `def join_path(*segments : String) : String`; rejects empty segments, maps each through `.chomp("/").lstrip("/")`, joins on `"/"`, and prepends `"/"` if the result doesn't already start with one. Pure, no side effects.
- `spec/unit_test/utils/utils_spec.cr` — the sibling spec for this same file; already tests `remove_start_slash` (line 4), `get_relative_path` (13), `any_to_bool` (76), `escape_glob_path` (106) but never `join_path`. Add the new `describe` block here.
- `spec/unit_test/utils/url_path_spec.cr:4` — ready copy-from template: a near-identical pure path-joining helper (`Noir::URLPath.join`) tested with plain `.should eq` String in/out assertions.

### Proposed change

Add a `describe "join_path"` block to `spec/unit_test/utils/utils_spec.cr` (the file already `require`s the function via `require "../../../src/utils/*"`, so no extra requires are needed). The expected values below were verified by running the real function against the current working tree:

```crystal
describe "join_path" do
  it "joins two segments" do
    join_path("/api", "users").should eq("/api/users")
  end

  it "collapses a double slash at the boundary" do
    join_path("/api", "/users").should eq("/api/users")
  end

  it "strips trailing slashes from segments" do
    join_path("api/", "/v1/").should eq("/api/v1")
  end

  it "skips empty segments" do
    join_path("", "users").should eq("/users")
  end

  it "joins three segments (variadic)" do
    join_path("a", "b", "c").should eq("/a/b/c")
  end

  it "always prepends a leading slash" do
    join_path("noslash").should eq("/noslash")
  end
end
```

**Pitfall to avoid:** test the *global* singular `join_path` from `src/utils/utils.cr` — do **not** confuse it with the global `join_paths` (plural, in `src/analyzer/analyzer.cr`) or with the unrelated private 2-arg `join_path` methods that the Dart analyzers (`src/analyzer/analyzers/dart/shelf.cr`, `angel3.cr`) define locally. The function under test is the variadic `*segments` one.

Verify with:

```sh
crystal spec spec/unit_test/utils/utils_spec.cr
```

### Why this is a good first issue

Small, self-contained, and test-only — it adds assertions to an existing spec file and touches no production code, so there is zero endpoint/param/callee detection (FP/FN) risk. The expected values are already verified, and the adjacent `url_path_spec.cr` provides a near-identical structure to copy from, so a newcomer can complete it confidently in one file.

---
🇰🇷 `src/utils/utils.cr`의 순수 헬퍼 `join_path`(형제 함수들은 모두 테스트가 있는데 이것만 누락)에 대해 `utils_spec.cr`에 `describe "join_path"` 블록을 추가하는 테스트 전용 작업입니다.

c0d33ngr:
PR to closes #2191 

## Description

This PR adds a dedicated unit test suite (`describe "join_path"`) to `utils_spec.cr` to directly verify the URL path segment composition of the global variadic `join_path` helper. 

While almost every other helper in `src/utils/utils.cr` already possessed explicit tests, `join_path` was the lone exception. This is a low-risk, test-only addition that brings full coverage consistency to the utility module.

### Changes
- **`spec/unit_test/utils/utils_spec.cr`**:
  - Appended test cases verifying double slash collapsing, trailing slash stripping, empty segment skipping, variadic joining, and leading slash enforcement.

### Verification
- Verified passing locally via `crystal spec spec/unit_test/utils/utils_spec.cr`.
- Ran `crystal tool format`.

### DIFF
--- a/spec/unit_test/utils/utils_spec.cr
+++ b/spec/unit_test/utils/utils_spec.cr
@@ -124,3 +124,29 @@ describe "escape_glob_path" do
     escape_glob_path("/x/{a}[b]*?\\c").should eq("/x/\\{a\\}\\[b\\]\\*\\?\\\\c")
   end
 end
+
+describe "join_path" do
+  it "joins two segments" do
+    join_path("/api", "users").should eq("/api/users")
+  end
+
+  it "collapses a double slash at the boundary" do
+    join_path("/api", "/users").should eq("/api/users")
+  end
+
+  it "strips trailing slashes from segments" do
+    join_path("api/", "/v1/").should eq("/api/v1")
+  end
+
+  it "skips empty segments" do
+    join_path("", "users").should eq("/users")
+  end
+
+  it "joins three segments (variadic)" do
+    join_path("a", "b", "c").should eq("/a/b/c")
+  end
+
+  it "always prepends a leading slash" do
+    join_path("noslash").should eq("/noslash")
+  end
+end


hahwul (approved):
LGTM

hahwul:
@c0d33ngr Thanks!

PR #2254, opened 2026-07-10
docs: rebuild the site around the project's own film-noir artwork

hahwul:
## Why

The docs site leaned on indigo `#818cf8`, the generic "AI indigo," while the project's own artwork is unmistakably film noir: `noir-wallpaper.jpg` is a monochrome city street at night with the wordmark knocked out of the frame, and the mascot Hak is a crane in a detective's trench coat whose only colour is a red crest. The name is French for black. The design was already in the repo; the site never used it.

This rebuilds the site around that, and takes the accessibility and correctness debt with it.

## Design

**One accent, derived not invented.** The red is Hak's crest, sampled at `#EA1623` and opened to `#EE2A22` so that both "red text on near-black" and "near-black label on a red fill" clear WCAG AA at 4.67:1 — the raw crest red only reaches 4.35:1. So primary CTAs are a red fill with a near-black label, never white on red (3.58:1). One radius scale, one z-index scale, no pure black or white. Third-party logos are grayscaled so Hwaro's brick red cannot leak in.

**The landing is a poster.** The wordmark baked into the photograph is the title; a restrained `h1` sits under it. The header is transparent only at scroll position zero.

**The docs reader** is a three-column layout (sidebar, 72ch prose, TOC). Only the branch holding the current page starts open, decided server-side so the tree never renders expanded and then snaps shut.

Typography moves from Inter to self-hosted Geist. Motion is native CSS plus IntersectionObserver: no scroll listeners, everything collapses under `prefers-reduced-motion`, and nothing is hidden unless `:root.js` is present — a page whose scripts fail to load stays readable.

## Nothing moves

All **136 URLs are byte-identical** to the previous build. I captured the old site's URL list before touching anything and diffed it at the end. Root files (`rss.xml`, `/ko/rss.xml`, `sitemap.xml`, `llms.txt`, `manifest.json`, `sw.js`, per-author feeds) are all preserved.

## Fixes carried along

- **Four orphaned pages.** The old sidebar silently stranded `usage/supported/{mobile,cli,ai_context_coverage}` and the whole `usage/cli_commands` section: they existed but nothing linked to them. All 61 pages are now reachable, each exactly once.
- **Missing alt text** on ten content images. Each now describes what the image actually shows, in both languages.
- **CLS.** Column-0 markdown images are emitted as `` with intrinsic dimensions, dropping CLS on `/get_started/running/` from 0.052 to 0.013.
- **Korean pages had no TOC** once `toc = true` moved into `[cascade]` — the `.ko.md` section indexes needed it too.
- **The release script would have silently stopped bumping the version.** `scripts/version_common.cr` grepped `class="hero-badge">v1.1.0<` out of both landing pages; that badge is gone. It now reads the version chip from the shared nav template.

## Verification

Measured, not asserted, against the built site:

| Check | Result |
| --- | --- |
| URL parity vs. old build | 136 / 136 identical |
| EN/KO doc parity (CI gate) | 66 / 66 |
| Contrast, both themes | 94 / 94 (incl. blog, authors, Korean) |
| Interactive (search, theme, drawer, copy, scroll-spy, reduced-motion) | 19 / 19 |
| i18n (search filtering, switcher round-trip, prev/next prefixes) | 6 / 6 |
| Sidebar expansion + position | 17 / 17 |
| Horizontal overflow, 19 pages x 4 widths | none |
| LCP (4x CPU throttle) | 808 ms, hero image |
| CLS (landing) | 0.0000 |
| `/noir` sub-path build | zero 404s, both languages |
| `hwaro doctor` / `tool validate` | clean |

## Notes for review

- The commits are a stacked redesign: `docs/` is replaced wholesale, so intermediate commits do not build on their own. The final state does.
- The blog and author templates are rebuilt in the new language (the aurora gradient is gone), but their content is untouched.
- 93 em-dashes remain in the existing prose. I did not rewrite your documentation copy; happy to in a follow-up.
- Four external links 404 and predate this branch: a hahwul.com post, `portswigger.net/burp`, an Apple simulator doc, and `src/miniparsers/go_route_extractor.cr`.
- `docs/AGENTS.md` records the traps that cost me time here: CommonMark ending an HTML block at a blank line, a `::after` painting above its siblings, `hwaro serve`'s first build using the config `base_url`, and `@media (max-width: 900px)` holding both the header collapse *and* the sidebar drawer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hahwul (commented):
## Review

Reviewed by exercising the built site, not only reading the diff. Overview, then findings ranked by severity.

**What it does:** replaces the docs theme (Inter/indigo → Geist/monochrome + one signal red), rebuilds every template, moves UI copy into `i18n/*.toml`, adds a language switcher, rebuilds blog/author pages, and repoints the release tooling. All 136 URLs are preserved.

The design work is sound and the palette reasoning is grounded (the 4.35:1 → 4.67:1 accent adjustment is the right call). The verification table checks out. But three real bugs slipped past those checks, all in the "everything renders, so it passes" blind spot.

---

### 🔴 Correctness

**1. `code.js` treats mermaid diagrams as code blocks** — `docs/static/js/code.js:18`

`document.querySelectorAll(".prose pre")` also matches `<pre class="mermaid">`, which the shortcode emits inside `.prose`. Every diagram gets wrapped in `.code-block` with a `CODE` language label and a Copy button. Clicking it copies the rendered SVG's node labels:

```
"Optimizer Detectors Condition Condition Condition Results Tags Techs …"
```

Affects 6 pages (`get_started/overview`, `get_started/ai_power`, `usage/output_formats/mermaid`, and their `/ko/` twins). Fix: `.prose pre:not(.mermaid)`.

**2. "Edit this page on GitHub" is broken on all 132 doc pages** — `docs/templates/page.html:16`

`{{ page.path }}` is empty — Hwaro exposes no source-file path (I probed `page.{path,relative_path,file,source,source_path,filename,slug}`; all empty). Every link resolves to:

```
https://github.com/owasp-noir/noir/edit/main/docs/content
```

i.e. a directory. It has to be derived from `page.url` + `page_language`, and section pages have no edit link at all.

**3. Korean pages link to the English author profile** — `blog_post.html:36`, `taxonomy.html:30`

`/ko/authors/hahwul/` exists and is fully Korean (`<html lang="ko">`, `전체 작성자`, links to `/ko/blog/…`), but Korean blog posts and the Korean authors index both link to `/authors/hahwul/`. The inline comment claiming "hwaro emits a single global /authors tree" is stale. Needs `lang_prefix`.

---

### 🟡 Regressions & consistency

**4. Sidebar is unusable without JavaScript** — `style.css:441,470`

Collapsed groups are `display: none` and the only way to open them is a JS click handler, so a no-JS visitor reaches **6 of 61** links. Before this PR every group was expanded server-side and JS only collapsed them. `<details>`/`<summary>` gives the same behaviour, keeps the server-side open state, works without JS, and deletes `wireCollapse`.

(The landing degrades correctly — 0 hidden `[data-reveal]` elements without JS. Good.)

**5. JS-side UI strings are not internationalized**

The PR moves "every UI string" into i18n tables, but five remain hardcoded English and appear on Korean pages:

- `code.js:59,64` — `Copy` / `Copied` / `Press Ctrl C`
- `search.js:75` — `No results for "…"`
- `theme.js:23,24` and `nav.html:44` — `Theme: auto` (aria-label + title)

---

### 🟠 Risk

**6. Mermaid is loaded from a floating CDN tag with no SRI** — `mermaid.js:69`

`https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.min.js` resolves to whatever the latest 10.x is. Pre-existing, but this file is rewritten here, so it is the moment to pin an exact version and add `integrity` + `crossorigin`.

**7. `check_version_consistency.cr` passes silently when a version is not found**

`compact_map` drops `nil`, so an unmatched file simply disappears from the comparison. I watched it print `docs/templates/partials/nav.html  Not found` and then `✅ All versions match`. Now that the nav template is the *only* docs file tracked, a markup change there would silently stop the release check from covering docs.

**8. Nit:** `.search-result span` sets `-webkit-line-clamp` with no standard `line-clamp` fallback.

---

### Not problems

- `[assets] fingerprint = true` does not rename `style.css`, so the direct `{{ base_url }}/css/style.css` reference is safe.
- `og.auto_image` does generate `og-images/*.png`.
- `search.js` escapes title, snippet, and href before `innerHTML`; the regex is escaped after HTML-escaping, so highlight matching is safe.
- `[markdown] safe = false` is required by the landing and the generated `tech-card` HTML; content is repo-trusted.

Fixes for 1–7 incoming.

hahwul:
## Review fixes pushed

All eight findings addressed, plus one more the fix surfaced.

| # | Finding | Fix |
| --- | --- | --- |
| 1 | mermaid rendered inside a code block, Copy copied SVG labels | `.prose pre:not(.mermaid)` |
| 2 | "Edit this page" pointed at `docs/content/` on all 132 pages | path rebuilt from `page.url` + language; **`docs/scripts/check_edit_links.sh`** asserts all 122 resolve |
| 3 | Korean pages linked to the English author profile | `lang_prefix` on both call sites |
| 4 | no-JS reader reached 6 of 61 sidebar links | `<details>`/`<summary>`; native disclosure, keyboard and a11y state; `wireCollapse` deleted |
| 5 | `Copy` / `No results for` / theme label hardcoded English | handed to the scripts via `window.NOIR_I18N`, English fallbacks |
| 6 | `mermaid@10` floating, no SRI | pinned `10.9.1` + `sha384` integrity + `crossorigin` |
| 7 | version check passed silently on an unreadable version | missing version is now a hard failure |
| 8 | `-webkit-line-clamp` only | left as-is; no standard `line-clamp` in the target browsers yet |

**A ninth, surfaced by fixing #1.** Once the diagrams rendered properly I could actually look at them in light mode: the overview diagram hardcodes `classDef highlight fill:#000`, which was safe while the docs were dark-only. Under light mode the node label is near-black and the `SourceCode` and `Report` nodes were black-on-black. They now carry the site accent with a near-black label — the same 4.67:1 pairing as the primary CTA.

**A tenth, self-inflicted.** `ui.theme` was itself a missing key, so the toggle's label read the literal string `ui.theme: auto`. That is precisely the footgun the i18n commit documented, and the unresolved-key scanner caught it. It now runs over all 132 built pages and reports zero.

### Verification after the fixes

11/11 review-fix assertions, 17/17 sidebar (expansion, position, keyboard, no-JS, drawer), 94/94 contrast in both themes, 19/19 interactive, 6/6 i18n, no horizontal overflow across 19 pages × 4 widths, LCP 808 ms, CLS 0.0000, 136/136 identical URLs, edit links 122/122, `doctor` / `validate` / `check_doc_parity` / `check_version_consistency` all clean.

Two notes for whoever touches this next, now recorded in `docs/AGENTS.md`:

- hwaro exposes **no** source-file path to templates. Anything that needs one has to derive it and be checked.
- `.prose pre` matches `<pre class="mermaid">`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2356, opened 2026-07-19
fix: perf-series follow-ups — recover dropped endpoints, bound memo, widen rescue

hahwul:
Follow-up to #2354. (Supersedes #2355, which GitHub auto-closed when #2354's branch was deleted on merge — same commits, now rebased onto `main`.)

## 1. Perf gates were dropping real endpoints

Two gates from the perf series trade false negatives for speed. For a tool whose job is enumerating attack surface that is the wrong direction, and both are silent — no warning, no diagnostic.

### `filter_redundant_generic_techs`

Dropped `js_http` / `go_http` / `python_starlette` whenever a same-language framework was detected. `techs` is a **flat, repo-wide list with no path granularity**, so a framework found *anywhere* suppresses the generic analyzer *everywhere*.

In a monorepo those belong to different applications. Measured on a 500-file Go corpus (400 Gin files + 100 `net/http` files):

| | unique URLs |
|---|---|
| before | 800 |
| after | **1000** |

+200 recovered, **0 lost**. The recovered set is exactly the raw `net/http` surface:

```
/debug/pprof/
/internal/secrets
/internal/shutdown
```

...i.e. the endpoints most worth finding. Same shape confirmed for Python — a standalone Starlette service beside a FastAPI one lost `/events/webhook` and `/events/internal/replay`.

Framework-vs-framework rules above are untouched and remain safe: both analyzers there target the same routing DSL. Scoping the stdlib rule correctly would need a per-tech path map from the detect loop (which does know the file) — a comment records that so it isn't re-added as-is.

### `zap_sites_tree`

Required a `*sites*` basename. ZAP export filenames are chosen by the user at export time, so an export saved as `zap_export.yaml` was dropped with no diagnostic — **0 endpoints, 7 after**. The expensive work is already gated in `detect` by the `includes?("Sites")` content guard, so the name gate only saved a substring scan.

## 2. Python memo bypassed its eviction cap

`extract_decorations_and_blueprints` wrote straight into the memo hashes, skipping the `DEFAULT_MAX_ENTRIES` eviction that `ExtractionResultCache.fetch` performs. It can't use `fetch` (that would parse twice), but the result was that the 4096 cap never applied to the **dominant** Python path — Flask, Quart and Sanic all use this combined entry point; the capped solo entries are the minority. On a large monorepo both memos grew one entry per file for the whole scan.

Extracted the capped insert as `ExtractionResultCache.store_capped`, so one parse still fills both stores but eviction applies.

## 3. Detector worker could hang the scan

The worker loop rescued only `File::NotFoundError`, unlike `parallel_analyze` which also has a broad `rescue e : Exception`. Any other exception from a detector or passive rule unwinds the worker to `ensure wg.done` and kills that fiber. Once every worker has died, the reader fiber blocks forever on `channel.send` — the scan hangs with no output and no crash. Mirrored `parallel_analyze`'s rescue.

## 4. Two comments describing behaviour the code doesn't have

- `source_fingerprint` claimed it mixes in `object_id`; it hashes content + bytesize only. Content-only is the *correct* choice — a recycled `object_id` would serve another file's table — the comment just described the opposite of the invariant.
- `@result_mutex` claimed fibers "can interleave on yield during Array reallocation". noir builds without `-Dpreview_mt`, so `Array#<<` has no yield point and cannot interleave today; the mutex is forward-looking for MT, not a fix for a live race.

## Left open deliberately

`@result_mutex` also **disagrees with `docs/content/development/analyzer_architecture`**, which documents the opposite convention (no mutex in per-file analyzers; synchronize at the `parallel_analyze` layer if MT lands), and the helpers are applied to only 7 engines — not the 5 Python analyzers this same series newly parallelized. Flagged in the comment rather than resolved: whether to remove it or apply it everywhere and update the docs is an architecture call, not a review fix.

Two other findings are **not** addressed here:
- `php_pure` is dropped by the same repo-wide rule when Laravel is present (pre-existing, not from this series) — a legacy `/upload.php` beside a Laravel app is lost.
- Parallelizing the Python walks makes endpoint ordering (and the reported `file:line` on duplicates) depend on worker count, i.e. on the host CPU count. Deterministic per machine, so CI won't see it; fixtures are far under the 128 channel capacity where divergence starts.

## Verification

| Check | Result |
|---|---|
| `crystal spec spec/unit_test` | 4087 / 0 failures |
| `crystal spec spec/functional_test` | 22383 / 0 failures |
| `ameba` | 1793 / 0 failures |
| `crystal tool format --check` | clean |

Re-run after the rebase onto `main`. Endpoint counts measured on this branch's binary.

### DIFF
--- a/spec/unit_test/analyzer/analyzer_selection_spec.cr
+++ b/spec/unit_test/analyzer/analyzer_selection_spec.cr
@@ -16,24 +16,20 @@ describe "filter_redundant_generic_techs" do
     filter_redundant_generic_techs(techs).should eq(techs)
   end
 
-  it "drops js_http when a JS framework analyzer is present" do
-    filter_redundant_generic_techs(["js_http", "js_express"]).should eq(["js_express"])
-    filter_redundant_generic_techs(["js_hono", "js_http", "js_fastify"]).should eq(["js_hono", "js_fastify"])
+  # Regression guard: a repo-wide framework hit must never suppress a
+  # generic stdlib analyzer. In a monorepo the two can belong to different
+  # applications (a standalone net/http admin listener beside a Gin API,
+  # a standalone Starlette service beside a FastAPI one), so dropping the
+  # generic analyzer silently loses real endpoints.
+  it "keeps go_http when a Go framework analyzer is also present" do
+    filter_redundant_generic_techs(["go_http", "go_gin"]).should eq(["go_http", "go_gin"])
   end
 
-  it "keeps js_http when no JS framework analyzer is present" do
-    filter_redundant_generic_techs(["js_http"]).should eq(["js_http"])
+  it "keeps js_http when a JS framework analyzer is also present" do
+    filter_redundant_generic_techs(["js_http", "js_express"]).should eq(["js_http", "js_express"])
   end
 
-  it "drops python_starlette when python_fastapi is present" do
-    filter_redundant_generic_techs(["python_fastapi", "python_starlette"]).should eq(["python_fastapi"])
-  end
-
-  it "keeps python_starlette alone" do
-    filter_redundant_generic_techs(["python_starlette"]).should eq(["python_starlette"])
-  end
-
-  it "drops go_http when a Go framework analyzer is present" do
-    filter_redundant_generic_techs(["go_http", "go_gin"]).should eq(["go_gin"])
+  it "keeps python_starlette when python_fastapi is also present" do
+    filter_redundant_generic_techs(["python_fastapi", "python_starlette"]).should eq(["python_fastapi", "python_starlette"])
   end
 end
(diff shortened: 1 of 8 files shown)


hahwul:
Pushed `e223ff76`: the fidelity sweep from #2354 was weaker than its commit message claimed, and I'd rather say so than leave it.

It only proves what the fixture corpus exercises. Deleting each `path_sensitive?` declaration and re-running showed Kamal's removal still **passed** — the Kamal fixture is `config/deploy.yml`, whose basename contains "deploy" and satisfies `applicable?` on its own, so nothing isolated the `/config/deploy` and `/.kamal/` clauses. That is precisely the failure mode the sweep exists to catch.

Added `DIRECTORY_GATE_PROBES` — layouts where the path matches a directory gate but the basename alone does not. Both load-bearing declarations are now guarded:

| declaration removed | before | after |
|---|---|---|
| `kamal` | passed | **1 failure** |
| `hasura` | 2 failures | 2 failures |

supabase and directus stay green without their declaration because `DETECTOR_PATH_SEGMENT_PROBES` already covers them, and strapi stays green because `schema_file?` matches the basename independently — so the memo is correct for it either way. Those three are belt-and-braces, not load-bearing.

PR #2366, opened 2026-07-20
fix: resolve Spring property placeholders in mapping paths

### LINKED ISSUE #2365: java_spring: Spring @RequestMapping property placeholders  are not resolved to real paths
**Describe the bug**
Noir’s Java Spring analyzer treats Spring configuration placeholders in @RequestMapping paths as literal URL segments instead of resolving them to their runtime values.

For example, Spring Boot’s BasicErrorController uses:

`@RequestMapping("${server.error.path:${error.path:/error}}")`
At runtime Spring resolves this to /error (or a custom value from application.properties / application.yml).

Noir currently reports a malformed path like `/${server.error.path:{error}} `(the optimizer partially rewrites the inner placeholder to {error} but still does not resolve the Spring property chain).

**To Reproduce**

1. Create a minimal Spring controller:

```
package com.test;
import org.springframework.http.ResponseEntity;
import org.springframework.stereotype.Controller;
import org.springframework.web.bind.annotation.RequestMapping;
@Controller
@RequestMapping("${server.error.path:${error.path:/error}}")
public class BasicErrorController {
    @RequestMapping
    public ResponseEntity<String> error() {
        return ResponseEntity.ok("error");
    }
}
```

2. Place it under a standard Maven/Gradle layout, e.g. `src/main/java/com/test/BasicErrorController.java`

3. Run Noir with the Java Spring tech enabled:

`noir -b /path/to/project -t java_spring -f json`

4. Observe the reported endpoint URL.

Actual output:

```
{
  "url": "/${server.error.path:{error}}",
  "method": "ANY"
}
```

**Expected behavior**
Noir should resolve Spring property placeholders using config when available, and otherwise use the declared defaults — matching Spring’s behavior.

For BasicErrorController with no custom config, the endpoint should be reported as:

```
{
  "url": "/error",
  "method": "ANY"
}
```
If `application.properties` contains `server.error.path=/custom-error`, the path should be `/custom-error`.

4ICH4:
## Summary

When scanning Java Spring services, Noir was treating `@RequestMapping` paths that use Spring configuration placeholders (e.g. `${server.error.path:${error.path:/error}}`, `${api.base.path:/api}`) as literal URL segments. Unresolved placeholders were reported as malformed paths like `/${server.error.path:{error}}`, and the optimizer could rewrite `${...}` segments into misleading `{name}` tokens instead of the runtime-resolved path.

## Changes

- `src/utils/spring_property_path.cr`: Add `SpringPropertyPath.resolve` to resolve nested `${property:default}` expressions from the inside out.
- `src/analyzer/analyzers/java/spring.cr`: Load project config from `application.properties` and flattened `application.yml` / `application.yaml`, cache values per project root, and apply placeholder resolution at every Java Spring endpoint URL emission site.
- `spec/functional_test/fixtures/java/spring/src/PlaceholderPathController.java`: Add regression fixtures for generic and error-path placeholder patterns.
- `spec/functional_test/fixtures/java/spring/resources/application.properties`: Provide `api.base.path` for config-driven resolution tests.
- `spec/unit_test/utils/spring_property_path_spec.cr`: Unit tests for nested defaults, config overrides, and inline defaults.
- `spec/functional_test/testers/java/spring_spec.cr`: Assert resolved endpoints `/error`, `/configured-api/items`, and `/configured-api/docs/guide`.

## Testing

- `crystal spec spec/unit_test/utils/spring_property_path_spec.cr`
- `crystal spec spec/functional_test/testers/java/spring_spec.cr`
Closes #2365

### DIFF
--- a/spec/functional_test/fixtures/java/spring/resources/application.properties
+++ b/spec/functional_test/fixtures/java/spring/resources/application.properties
@@ -1,0 +1,1 @@
+api.base.path=/configured-api
--- a/spec/functional_test/fixtures/java/spring/src/PlaceholderPathController.java
+++ b/spec/functional_test/fixtures/java/spring/src/PlaceholderPathController.java
@@ -1,0 +1,32 @@
+package com.test;
+
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("${api.base.path:/api}")
+public class PlaceholderPathController {
+
+    @GetMapping("/items")
+    public String listItems() {
+        return "items";
+    }
+
+    @GetMapping("${app.docs.path:/docs}/guide")
+    public String docsGuide() {
+        return "guide";
+    }
+}
+
+@Controller
+@RequestMapping("${server.error.path:${error.path:/error}}")
+class BasicErrorController {
+
+    @RequestMapping
+    public ResponseEntity<String> error() {
+        return ResponseEntity.ok("error");
+    }
+}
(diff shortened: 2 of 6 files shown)


hahwul (approved):
LGTM! Great work

PR #2101, opened 2026-06-15
Use idiomatic .empty? checks in send_proxy

gitrlawton:
Replaces `.size > 0` with `.empty?` in `src/deliver/send_proxy.cr` to match the idiomatic Crystal style used consistently throughout the rest of the codebase.

- `endpoint.params.size > 0` changed to `unless endpoint.params.empty?`
- `endpoint_hash["json"].size > 0` changed to `!endpoint_hash["json"].empty?`

Fixes #1089

Issue #1121, opened 2026-03-22
Use idiomatic .empty? instead of .size > 0 across the codebase


hahwul:
## Description

The codebase extensively uses `.size > 0`, `.size == 0`, and `.size != 0` for collection/string emptiness checks instead of Crystal's idiomatic `.empty?` / `!.empty?`.

**Note:** #1089 covers only `src/deliver/send_proxy.cr`. This issue tracks the remaining ~230+ instances across the entire codebase.

### Affected files (sampling)

- `src/analyzer/analyzers/ruby/rails.cr` — `stripped_line.size > 0`
- `src/analyzer/analyzers/typescript/tanstack_router.cr` — 10+ instances of `match.size > 0`
- `src/analyzer/analyzers/crystal/grip.cr` — `current_scopes.size > 0`
- `src/analyzer/analyzers/java/spring.cr` — `.size > 0` and `.size != 0`
- `src/analyzer/analyzers/go/common.cr`, `gin.cr`, `echo.cr`, `fiber.cr`, `mux.cr`, `beego.cr`, `gf.cr` — many instances
- `src/analyzer/analyzers/python/python.cr`, `django.cr`, `fastapi.cr`, `flask.cr` — `.size != 0`
- `src/analyzer/analyzers/javascript/express.cr`, `restify.cr` — many instances
- `src/miniparsers/js_route_extractor.cr` — 30+ instances
- `src/optimizer/llm_optimizer.cr` — 6 instances

### Expected

```crystal
# Before
if items.size > 0
if items.size == 0
if items.size != 0

# After
unless items.empty?
if items.empty?
unless items.empty?
```

This can be tackled incrementally per directory or analyzer.

PR #2376, opened 2026-07-23
refactor(output,deliver): use idiomatic .empty? for collection emptiness (#1121)

MohammedAnasNathani:
## Summary

Refs #1121.

Convert `.size > 0` to `!.empty?` in three files inside the **output** and **deliver** layers, all of which are guarding real `Array` / `Hash` collections (not `String` / `Tuple` results from `String#scan`, which lack `.empty?`).

| File | Sites | Notes |
|---|---|---|
| `src/output_builder/diff.cr` | 4 | `result[:added]`, `result[:removed]`, `result[:changed]`, `endpoints.as_a` |
| `src/output_builder/common.cr` | 9 | `header_params`, `cookie_params`, `baked[:path_param]`, `extra_params`, `flag_params`, `argument_params`, `env_params`, `tags` (the 8th & 9th) |
| `src/deliver/send_req.cr` | 2 | `endpoint.params`, `endpoint_hash["json"]` |

Each guard is a pure collection-empty check; the conversion is semantically identical for non-negative sizes (`.empty?` is the Crystal-idiomatic equivalent).

## Why this subset and not the rest of the codebase

The remaining `.size > 0` / `.size == 0` / `.size != 0` sites in `src/` are almost all inside `String#scan` callbacks where the yielded receiver is a Crystal `Tuple` (1-based indexing, fixed size, no `.empty?` method):

```crystal
# src/analyzer/analyzers/javascript/restify.cr (not touched here)
handler_body.scan(/req\.query\.(\w+)/) do |param_match|
  if param_match.size > 0           # param_match is a Tuple(String)
    endpoint.push_param(Param.new(param_match[1], "", "query"))
  end
end
```

- One capture group → `Tuple(String)`
- Two capture groups → `Tuple(String, String)`
- Tuples don't define `.empty?`, so this conversion is not possible without changing the regex shape

Splitting the regex-yielded sites into a separate pass keeps this PR reviewable. The track issue (#1121) makes that easy to do incrementally.

## Already done by others

- `src/deliver/send_proxy.cr` was converted in #2101.
- This PR picks up the remaining deliver/output surface that #2101 didn't reach.

## Regression coverage

New test in `spec/unit_test/output_builder/diff_spec.cr`:

```crystal
it "produces three empty buckets for identical input" do
  ...
  result = builder.diff(endpoints, endpoints)
  result[:added].empty?.should be_true
  result[:removed].empty?.should be_true
  result[:changed].empty?.should be_true
end
```

This locks in the empty-bucket behavior that `print` / `generate_toml_from_diff` now depend on (every bucket empty → no section header rendered). The conversion is mechanical, so the existing `diff()` tests already exercise the populated-bucket paths.

## Files changed

- `src/output_builder/diff.cr` — 4 sites
- `src/output_builder/common.cr` — 9 sites
- `src/deliver/send_req.cr` — 2 sites
- `spec/unit_test/output_builder/diff_spec.cr` — 1 new test

15 collection-empty sites converted + 1 regression test, all in 4 files.

### DIFF
--- a/spec/unit_test/output_builder/diff_spec.cr
+++ b/spec/unit_test/output_builder/diff_spec.cr
@@ -1,6 +1,7 @@
 require "../../spec_helper"
 require "../../../src/output_builder/diff"
 require "../../../src/models/endpoint"
+require "../../../src/models/noir"
 require "../../../src/utils/utils"
 
 describe "OutputBuilderDiff" do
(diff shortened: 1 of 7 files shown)


hahwul (approved):
LGTM

hahwul:
Hi @MohammedAnasNathani 
Thanks for contributing and Nice work! It's merged :)

PR #2453, opened 2026-08-01
refactor: use idiomatic zero-size predicates

### LINKED ISSUE #2450: Use idiomatic .empty? instead of .size == 0 / .size != 0
Earlier sweeps replaced `.size > 0` with `.empty?` across the codebase, but the
mirror-image forms `.size == 0` and `.size != 0` are still around. Every
receiver below is a `String` or an `Array`, so `.empty?` / `!...empty?` is a
drop-in replacement.

```console
$ grep -rn '\.size == 0\|\.size != 0' src --include="*.cr" | grep -v ext/
src/utils/tnetstring.cr:89
src/analyzer/analyzers/python/flask.cr:416, 1480, 1500, 1503
src/analyzer/analyzers/python/django.cr:1850, 1854, 1881, 1885
src/analyzer/analyzers/python/quart.cr:794, 813, 816
src/analyzer/engines/python_engine.cr:230, 393, 638, 639
src/analyzer/engines/go_engine.cr:471
```

17 sites in 7 files.

## What to do

Rewrite each one:

```crystal
# before
if matches.size == 0
next if _match.size == 0
if @type.size != 0

# after
if matches.empty?
next if _match.empty?
unless @type.empty?
```

For the `!= 0` sites, use whichever of `if !x.empty?` / `unless x.empty?` reads
better in context — the codebase prefers `unless` for the plain negated form.
One site is already inside an `unless`, so watch the polarity there:

```crystal
# src/utils/tnetstring.cr:89 — before
raise ParseError.new("null payload must be empty") unless payload.size == 0
# after
raise ParseError.new("null payload must be empty") unless payload.empty?
```

Nothing here changes behavior — `String#empty?` and `Array#empty?` are exactly
`size == 0`.

## Verify

```bash
just check
crystal spec spec/unit_test
```

Vam-si-krish:
## Summary

- replace the 16 String/Array zero-size comparisons with `empty?` and readable `unless` forms
- use `size.zero?` for the remaining `_match` receiver, which is `Regex::MatchData` and does not implement `empty?`
- preserve every branch polarity and remove all `.size == 0` / `.size != 0` forms from `src/`

The `Regex::MatchData` distinction corrects one premise in the issue while keeping its intended behavior and idiomatic zero check. Applying `empty?` at that site fails Crystal compilation.

Closes #2450

## Validation

- `docker build --target builder -t noir-pr2450-check .` — full Crystal 1.21 static build passed
- `docker run --rm noir-pr2450-check crystal tool format --check`
- `docker run --rm noir-pr2450-check crystal spec spec/unit_test`
- `docker run --rm noir-pr2450-check crystal spec spec/functional_test`
- confirmed no `.size == 0` or `.size != 0` remains under `src/`

AI-assisted contribution; I reviewed every changed branch and used the compiler to validate receiver-specific APIs.

### DIFF
--- a/src/analyzer/analyzers/python/django.cr
+++ b/src/analyzer/analyzers/python/django.cr
@@ -1847,11 +1847,11 @@ module Analyzer::Python
         REQUEST_PARAM_FIELD_PATTERNS.each do |field_pattern|
           field_name, field_methods, param_type, bracket_re, get_re = field_pattern
           matches = line.scan(bracket_re)
-          if matches.size == 0
+          if matches.empty?
             matches = line.scan(get_re)
           end
 
-          if matches.size != 0
+          unless matches.empty?
             matches.each do |match|
               next if match.size != 2
               param_name = match[1]
@@ -1878,11 +1878,11 @@ module Analyzer::Python
 
       if line.includes? "form.cleaned_data"
         matches = line.scan(/form\.cleaned_data\[[rf]?['"]([^'"]*)['"]\]/)
-        if matches.size == 0
+        if matches.empty?
           matches = line.scan(/form\.cleaned_data\.get\([rf]?['"]([^'"]*)['"]/)
         end
 
-        if matches.size != 0
+        unless matches.empty?
           matches.each do |match|
             next if match.size != 2
             suspicious_params << Param.new(match[1], "", "form")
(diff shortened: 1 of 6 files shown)


hahwul (approved):
LGTM!

hahwul:
Hi @Vam-si-krish
Thanks for the contribution! Nice code.
There was a CI failure, so I fixed it a bit and merged :)

PR #2458, opened 2026-08-02
test: add Scala lexer unit specs

### LINKED ISSUE #2452: Add unit specs for Noir::ScalaLexer
`spec/unit_test/minilexer/` covers only two of the five mini-lexers:

```console
$ ls src/minilexers/
csharp_lexer.cr  golang.cr  js_lexer.cr  php_lexer.cr  sample.cr  scala_lexer.cr
$ ls spec/unit_test/minilexer/
golang_spec.cr  js_lexer_spec.cr
```

`Noir::ScalaLexer` (`src/minilexers/scala_lexer.cr`, 355 lines) has no direct
spec at all — it is only exercised indirectly through the Scala analyzer
fixtures, so a regression in the lexer surfaces as a confusing endpoint diff
rather than a pinpointed failure.

## What to do

Add `spec/unit_test/minilexer/scala_lexer_spec.cr`. Use
`spec/unit_test/minilexer/js_lexer_spec.cr` as the structural template
(`require "spec"`, `require "../../../src/minilexers/scala_lexer"`, one
`describe` per public method).

The public surface, and what is worth asserting:

| Method | Suggested cases |
| --- | --- |
| `masked` | strings, `//` and nested `/* /* */ */` comments, `"""…"""` bodies and `'x'` char literals are all blanked; result is the same length as the source and keeps `\n` |
| `code` | same as above **except** regular `"…"` string literals are preserved — this is the view the Scala analyzers read route paths from |
| `masked_lines` / `code_lines` | line count matches `source.lines.size` |
| `matching_delimiter(open_pos)` | `(`/`[`/`{` pairs; a brace inside a string or comment is ignored; returns `nil` on an unbalanced opener and on a non-delimiter position |
| `statement_end(start_pos)` | stops at a top-level `;`, ignores `;` nested inside `(...)`/`[...]`/`{...}`, returns the source size when there is none |
| `in_code?(pos)` | true for code positions, false inside a comment or string span |
| `tokens` | `kind` / `value` / `start` / `end` / `line` on a small snippet; `line` is 1-based |

A good regression case to include, since it is the reason the lexer exists: a
route-shaped call inside a triple-quoted string or a multi-line block comment
must be blanked in `masked` and must not appear as code.

```scala
val doc = """
  path("phantom") { get { complete("x") } }
"""
path("real") { get { complete("y") } }
```

## Verify

```bash
crystal spec spec/unit_test/minilexer/scala_lexer_spec.cr
just check
```

Only the new spec file should change — this issue adds tests, it does not
change `scala_lexer.cr`. If a case you wrote fails, please describe it in the
PR rather than "fixing" the lexer to match the test; the behavior may be
intentional.

Vam-si-krish:
## Summary

Adds unit specs for `Noir::ScalaLexer` covering masked/code views, nested comments and strings, delimiter matching, statement boundaries, and token metadata. The tests exercise the existing lexer contract without changing implementation behavior.

Closes #2452

## Validation

- `crystal spec spec/unit_test/minilexer/scala_lexer_spec.cr` in `crystallang/crystal:1.19.0-alpine` (8 examples, 0 failures)
- `crystal tool format --check` in the same container (pass)

The repository-wide Ameba/unit/functional check was also started in the same container and was still compiling its first full scan when this PR was opened; no implementation files are changed.

### DIFF
--- a/spec/unit_test/minilexer/scala_lexer_spec.cr
+++ b/spec/unit_test/minilexer/scala_lexer_spec.cr
@@ -24,8 +24,8 @@ describe Noir::ScalaLexer do
       masked.should_not contain("real")
       code.should_not contain("phantom")
       code.should contain("path(\"real\")")
-      lexer.in_code?(source.index("phantom").not_nil!.to_i).should be_false
-      lexer.in_code?(source.index("path(\"real\")").not_nil!.to_i).should be_true
+      lexer.in_code?(source.index!("phantom")).should be_false
+      lexer.in_code?(source.index!("path(\"real\")")).should be_true
     end
 
     it "keeps masked and code line counts aligned with the source" do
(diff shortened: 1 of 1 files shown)


hahwul (approved):
LGTM!

hahwul:
Hi @Vam-si-krish 
Nice code! thanks for contributing
(I fix lint CI error)

PR #2539, opened 2026-08-13
refactor(output): drive the terminal field list from a table

hahwul:
## Summary

`OutputBuilder::Common#print` was ~206 lines and 66 branch points, made of twenty near-identical "if the collection is non-empty, append a labelled row" blocks. Adding a field meant copying six lines for the twenty-first time.

| | before | after |
|---|---|---|
| `print` lines | 206 | **158** |
| branch points in the method | 66 | **39** |
| incl. the two new helpers | 66 | 42 |

The file total is unchanged (277): the ~48 lines left the method and came back as two tables and two comment-heavy helpers. The win is not line count — it is that a new field is now a table row.

## It was two shapes, not one

The twenty blocks turned out to be two, so there are two appenders rather than one loop:

- `append_field(buf, label, value, color)` — a `\n  ○ <label>: <value>` row; `color : Symbol?`, nil meaning "print raw".
- `append_tree_field(buf, label, entries, color)` — a `○ <label>: ` header with entries hanging below as `├──`/`└──`.

Fed by two constant tables (`PARAM_TREE_FIELDS`, `PARAM_NAME_FIELDS`) plus a local tuple for the six ai_context categories, which replaces a 6-clause `||` chain and six hand-written `if` lines. Every remaining row — `path`, mobile metadata, both `body` forms, `tags`, `tech`, `file`, `callees` — is one call to one appender, and each row's original comment travels with it.

## Ten near-misses between the "identical" blocks — all preserved, all documented in place

This is why the table was built from the real code rather than a template:

1. `tags` joins on `" "`; every other multi-value row joins on `", "`.
2. `tags` is `:light_magenta`; the other joined rows are `:cyan`.
3. The form-body tree omits the `[unresolved]` marker its visually identical header/cookie siblings append.
4. `headers` separates name/value with `": "`; `cookies` and form-body use `"="`.
5. `file` is the only row printed with no colour at all.
6. `ai_context:` is the only header with no trailing space after the colon.
7. `tech` uses `:light_blue` and is gated on option + truthiness rather than `.empty?`.
8. The ai_context feature name and printed label disagree for one row: the flag is `--ai-context callee`, the block prints `callees`.
9. The tree headers emit a trailing space that nothing follows.
10. `extras` is emitted from two different sources, so an endpoint carrying both prints two `○ extras:` rows.

## Real defects found, deliberately NOT fixed

- **Two `MOBILE_METADATA_KEYS` rows can never print**: `metadata["query"]` and `metadata["extras"]` are written by no analyzer in `src/`. Left in place, in order.
- `if details.code_paths && !details.code_paths.empty?` — `code_paths` is a non-nil `Array(PathInfo)`, so the first half is always true. Left verbatim.

## Test plan

**The 667-project JSON A/B does not cover this method at all** — it compares endpoint JSON, and this renders terminal text. The oracle is the rendered output.

- **108 captures (54 cases × 2 colour modes) byte-identical**, taken against the unmodified tree. Colour captures run stdout on a real pty, because `@is_color` is `… && STDOUT.tty?` — a plain pipe silently disables the very thing under test.
- Independently re-verified after rebasing onto current `main`: **24 captures over 12 fixtures, 73,028 bytes, byte-identical**, with ANSI escapes confirmed present in every colour capture.
- Coverage observed in the captures: all 20 `○` labels the code can emit, both tree prefixes, both `name: value` / `name=value` forms, `[unresolved]` markers, all four mobile protocol prefixes, `[server-action]`, `[websocket]`, all five ai_context categories that fire, all four status-colour branches plus the `[error]` no-server branch, and passive scan with and without findings. The only unreachable rows are the two dead metadata keys above.
- `crystal spec spec/unit_test` — 4,760 examples, 0 failures. `crystal spec spec/functional_test` — 22,699 examples, 0 failures.
- `ameba`, `crystal tool format --check`, repo-wide `typos` — clean.

### DIFF
--- a/src/output_builder/common.cr
+++ b/src/output_builder/common.cr
@@ -25,6 +25,30 @@ class OutputBuilderCommon < OutputBuilder
                           "read_permission", "write_permission", "grant_uri_permissions",
                           "path_permissions", "extras"]
 
+  # Parameter kinds drawn as an indented tree under a `○ <label>: ` header,
+  # in emission order: `{param_type, label, name/value separator}`. Both are
+  # green and both mark params the analyzer could not resolve; the separator
+  # follows the wire syntax of the thing being printed (`Name: value` for a
+  # header, `name=value` for a cookie).
+  PARAM_TREE_FIELDS = [
+    {"header", "headers", ": "},
+    {"cookie", "cookies", "="},
+  ]
+
+  # Parameter kinds drawn as one cyan comma-joined line of names, in emission
+  # order: `{param_type, label}`.
+  PARAM_NAME_FIELDS = [
+    # Intent extras (Bundle inputs, not part of the URI) read by a mobile
+    # handler. Query params bake into the URL, so only "extra" is listed.
+    {"extra", "extras"},
+    # CLI inputs (protocol "cli"): named flags/options, positional arguments,
+    # and consumed environment variables. HTTP endpoints have no params of
+    # these types, so these rows only render for CLI endpoints.
+    {"flag", "flags"},
+    {"argument", "arguments"},
+    {"env", "env"},
+  ]
+
   @ai_context_features : Set(String)? = nil
 
   # The plain report is the only format that frames itself — a heading over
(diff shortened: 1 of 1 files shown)


hahwul:
Squash-merged as `155db7f3` — the merge API call returned a 502 from GitHub after the commit landed, so this PR record was left open. Verified: `src/output_builder/common.cr` on `main` is byte-identical to this branch's version. Closing manually.

PR #2567, opened 2026-08-15
feat(analyzer): detect HTTP QUERY routes in Fastify

### LINKED ISSUE #2544: feat(analyzer): detect HTTP QUERY routes in Fastify
Depends on foundation #2540

## Upstream status

**Supported via the documented extension point, not in the default verb set.** fastify/fastify#6807 was closed with the maintainers preferring to wait before putting `QUERY` in core; the recommended path today is `addHttpMethod`, which then generates the route shorthand:

```js
fastify.addHttpMethod('QUERY', { hasBody: true })

fastify.query('/search', handler)                              // shorthand, now available
fastify.route({ method: 'QUERY', url: '/search', handler })    // full declaration
```

There is also the `fastify-http-query` plugin from a Fastify maintainer, which does the same registration.

## Route shapes to detect

- `fastify.route({ method: 'QUERY', ... })` and the array form `method: ['GET', 'QUERY']`
- `fastify.query('/path', ...)` after an `addHttpMethod('QUERY')` call
- The same shapes inside a plugin/encapsulated context

## Noir changes

- `src/analyzer/analyzers/javascript/fastify.cr:541` — `HTTP_METHODS = %w[get post put delete patch options head]`; add `query`.
- Same file `:188` — the second inline verb list.
- The `method:` string/array form should already flow through once the verb table accepts `QUERY`; verify.
- Optional but preferred: only treat `fastify.query(...)` as a route when an `addHttpMethod('QUERY'` call is present in the project, since without it the shorthand does not exist. The `method: 'QUERY'` form needs no such gate.

## Tests

- [ ] Fixture with `addHttpMethod('QUERY', { hasBody: true })` + both route shapes
- [ ] Fixture without the `addHttpMethod` call, asserting no phantom `QUERY` endpoint from a stray `.query(` call

---
🇰🇷 요약: Fastify는 `addHttpMethod('QUERY')` 이후 `fastify.query()` / `method: 'QUERY'`로 라우트를 선언 — 해당 형태를 탐지하되 등록이 없으면 오탐하지 않도록 게이트.

hahwul:
## Description

Implements detection for HTTP `QUERY` method routes in Fastify (RFC 10008).

### Supported Route Shapes
1. **Explicit Route Configurations**:
   - `fastify.route({ method: 'QUERY', url: '/search', ... })`
   - Array form: `fastify.route({ method: ['GET', 'QUERY'], url: '/items/:id', ... })` and `methods: [...]`
   - Always detected without gate since explicit method declaration is unambiguous.

2. **Shorthand Method Calls**:
   - `fastify.query('/search', handler)` (and `app`, `server`, `instance` receivers)
   - Plugin encapsulated routes with prefix support (`fastify.register(plugin, { prefix: '/api' })`)
   - **Gated**: only active when `addHttpMethod('QUERY' ...)` or the `fastify-http-query` plugin is present in the project to prevent false positives from stray `.query(` calls (e.g. database client queries).

3. **Detector & Extractor Updates**:
   - Updated Fastify detector signal to recognize `.query(` and `.route(`.
   - Updated `FASTIFY_RECEIVER_MARKER` in `js_route_extractor.cr` to include `query`.

### Tests Added
- Fixture and tester for Fastify with `addHttpMethod('QUERY')` + both route shapes and plugins.
- Fixture and tester for Fastify without `addHttpMethod` asserting no phantom `QUERY` endpoints from stray `.query(` calls.
- Updated Fastify detector unit test.

Closes #2544


hahwul:
Rebased onto main after the JS-cluster merges (#2569/#2570/#2568). Maintainer change per review: the shared JS parser now recognizes `.query(` for any receiver (ungated, unprefixed), which bypassed this PR's `addHttpMethod('QUERY')` gate and duplicated plugin-prefixed routes — the analyzer now skips shared-parser QUERY endpoints so the gated Fastify pass owns all QUERY emission. Verified: both new fastify specs fail on the combined tree without this line and pass with it.

PR #2653, opened 2026-08-19
fix(minilexer): keep Token#to_s side-effect free

### LINKED ISSUE #2646: `Token#to_s` rewrites the token’s own `value` as a side effect
`Token#to_s` in `src/models/minilexer/token.cr` escapes a newline or tab for
display by *assigning to the token's own `@value`*:

```crystal
def to_s
  if @value == "\n"
    @value = "\\n"
  elsif @value == "\t"
    @value = "\\t"
  end
  "#{@type} '#{@value}'"
end
```

So printing a token rewrites it. A `:NEWLINE` token that goes through `to_s`
once carries the two-character string `\n` from then on, and any later
`token.value` read gets the escaped form instead of the real one. `to_s` is a
formatting method; it should not be able to change what it is formatting.

Nothing is broken in production today — the only caller is the debug printer
`MiniLexer#trace` (`src/models/minilexer/minilexer.cr:81`), which prints tokens
and discards them — which is exactly why this is a safe first change.

```console
$ grep -rn 'token.to_s' src/
src/models/minilexer/minilexer.cr:81:      puts token.to_s
```

## What to do

Compute the display form in a local and leave `@value` alone. The idiomatic
Crystal shape also writes into the `IO` rather than building a `String`:

```crystal
def to_s(io : IO) : Nil
  display = case @value
            when "\n" then "\\n"
            when "\t" then "\\t"
            else           @value
            end
  io << @type << " '" << display << "'"
end
```

Keep the behavior exactly as it is: only a value that is *entirely* `"\n"` or
`"\t"` is escaped, and the output stays `type 'value'`. Do not switch to
`gsub` — that would also escape newlines embedded in a longer token, which is a
different output.

`token.to_s` (no argument) keeps working through `Object#to_s`, so
`MiniLexer#trace` needs no change.

The four existing `describe "to_s"` cases in
`spec/unit_test/models/minilexer/token_spec.cr` must pass unchanged. Adding one
that pins down the fix is welcome:

```crystal
it "does not mutate the token's value" do
  token = Token.new(:newline, "\n", 0)
  token.to_s
  token.value.should eq("\n")
end
```

## Verify

```bash
just check
crystal spec spec/unit_test/models/minilexer/token_spec.cr
crystal spec spec/unit_test
```

landon-personal:
## Summary

- render newline and tab values without rewriting `@value`
- use Crystal's IO-based `to_s` overload while preserving output
- add a regression test for both escaped whitespace values

## Why

`Token#to_s` escaped display output by assigning back to the token's value, so formatting changed what later readers observed.

## Validation

- `crystal tool format --check`
- Ameba: 2,276 inspected, 0 failures
- focused specs: 11 examples, 0 failures
- unit suite: 5,556 examples, 0 failures, 8 pending
- functional suite: 24,415 examples, 0 failures

Fixes #2646

### DIFF
--- a/spec/unit_test/models/minilexer/token_spec.cr
+++ b/spec/unit_test/models/minilexer/token_spec.cr
@@ -59,6 +59,16 @@ describe "Token" do
       str.should contain("\\t")
       str.should_not contain("\t")
     end
+
+    it "does not mutate the token's value" do
+      ["\n", "\t"].each do |value|
+        token = Token.new(:whitespace, value, 0)
+
+        token.to_s
+
+        token.value.should eq(value)
+      end
+    end
   end
 
   describe "property setters" do
--- a/src/models/minilexer/token.cr
+++ b/src/models/minilexer/token.cr
@@ -17,12 +17,12 @@ class Token
     @type == type
   end
 
-  def to_s
-    if @value == "\n"
-      @value = "\\n"
-    elsif @value == "\t"
-      @value = "\\t"
-    end
-    "#{@type} '#{@value}'"
+  def to_s(io : IO) : Nil
+    display = case @value
+              when "\n" then "\\n"
+              when "\t" then "\\t"
+              else           @value
+              end
+    io << @type << " '" << display << "'"
   end
 end


hahwul (approved):
LGTM

hahwul:
Hi @landon-personal 
Thanks for contributing and Nice work! Merged

PR #2667, opened 2026-08-28
fix(tagger): ignore a trailing slash on a guarded prefix

### LINKED ISSUE #2650: Add a spec for the shared `PrefixScope.prefix_covers?` tagger helper
`src/tagger/framework_taggers/prefix_scope.cr` is a 23-line module holding one
pure function, `prefix_covers?`, and it has no spec of its own:

```console
$ grep -rn 'prefix_covers?' src --include="*.cr"
src/tagger/framework_taggers/prefix_scope.cr:19:  def prefix_covers?(prefix : String, url : String) : Bool
src/tagger/framework_taggers/go/go_auth.cr:202:      next unless prefix_covers?(scope[:prefix], url)
src/tagger/framework_taggers/go/go_security.cr:152:      next unless prefix_covers?(scope[:prefix], endpoint.url)
src/tagger/framework_taggers/javascript/express_auth.cr:243:      if prefix_covers?(rule[:prefix], url)
src/tagger/framework_taggers/kotlin/ktor_auth.cr:252:        return scope[:description] if prefix_covers?(scope[:prefix], url)
$ ls spec/unit_test/tagger/framework_taggers/ | grep prefix
$
```

Four taggers across three languages decide "does this scoped middleware guard
this endpoint?" through it, so its behavior is only ever asserted indirectly, one
framework at a time. The module's own comment explains why the naive
`url.starts_with?(prefix)` is wrong here — `/admin` must not be reported as
guarding `/administration/report` — and that is exactly the property worth
pinning down directly.

## What to do

Add `spec/unit_test/tagger/framework_taggers/prefix_scope_spec.cr`. The function
is pure `(String, String) -> Bool`: no IO, no options, no tagger state, so the
spec needs nothing but the module.

```crystal
require "../../../spec_helper"
require "../../../../src/tagger/framework_taggers/prefix_scope"

describe PrefixScope do
  it "covers the prefix itself" do
    PrefixScope.prefix_covers?("/web", "/web").should be_true
  end

  it "covers a child path" do
    PrefixScope.prefix_covers?("/web", "/web/x").should be_true
  end

  it "does not cover a longer sibling segment" do
    PrefixScope.prefix_covers?("/admin", "/administration/report").should be_false
    PrefixScope.prefix_covers?("/web", "/website").should be_false
  end

  it "treats the root prefix as covering everything" do
    PrefixScope.prefix_covers?("/", "/anything/at/all").should be_true
  end
end
```

Worth covering too, since taggers pass whatever a route group yielded: a prefix
with a trailing slash (`"/web/"`), an empty URL, and a URL that differs only by
case. Assert whatever the current implementation actually does — this issue
pins behavior down, it does not change it. If a case looks wrong to you, say so
in the PR instead of "fixing" `prefix_covers?`.

The module is `extend self` and is also `include`d into three taggers, so both
`PrefixScope.prefix_covers?(...)` and a `include PrefixScope` inside the spec
work; the qualified form above is clearer.

## Verify

```bash
just check
crystal spec spec/unit_test/tagger/framework_taggers/prefix_scope_spec.cr
crystal spec spec/unit_test
```

chei-l:
Closes #2650

### Description
Adds `spec/unit_test/tagger/framework_taggers/prefix_scope_spec.cr` to directly verify `PrefixScope.prefix_covers?`.

Covers:
- Covering the prefix itself
- Child path containment
- Rejection of longer sibling segments (e.g. `/admin` vs `/administration/report`, `/web` vs `/website`)
- Root prefix `/` covering all endpoints (including empty string)
- Rejection of unrelated paths
- Trailing slash behavior on prefixes
- Empty URL handling
- Case sensitivity

### Follow-up fix (maintainer)
Writing the trailing-slash case surfaced a real bug, so the second commit fixes it: `prefix_covers?` compared the prefix verbatim, so `app.use('/admin/', requireAuth)` — which in Express mounts exactly what `'/admin'` does — matched neither `/admin` nor `/admin/users`, and every endpoint behind that middleware was reported as unguarded. Trailing slashes are now stripped before the segment-boundary compare, and the spec asserts the corrected behavior (plus a url-side trailing slash and an empty prefix). Go and Ktor normalize their prefixes upstream (`join_prefix` / `normalize_prefix`), so only Express changes behavior.

### Verification
- `just check` passed
- `crystal spec spec/unit_test` passed (5565 examples, 0 failures)
- `crystal spec spec/functional_test` passed (24429 examples, 0 failures)

### DIFF
--- a/spec/unit_test/tagger/framework_taggers/prefix_scope_spec.cr
+++ b/spec/unit_test/tagger/framework_taggers/prefix_scope_spec.cr
@@ -1,0 +1,42 @@
+require "../../../spec_helper"
+require "../../../../src/tagger/framework_taggers/prefix_scope"
+
+describe PrefixScope do
+  describe ".prefix_covers?" do
+    it "covers the prefix itself" do
+      PrefixScope.prefix_covers?("/web", "/web").should be_true
+    end
+
+    it "covers a child path" do
+      PrefixScope.prefix_covers?("/web", "/web/x").should be_true
+    end
+
+    it "does not cover a longer sibling segment" do
+      PrefixScope.prefix_covers?("/admin", "/administration/report").should be_false
+      PrefixScope.prefix_covers?("/web", "/website").should be_false
+    end
+
+    it "treats the root prefix as covering everything" do
+      PrefixScope.prefix_covers?("/", "/anything/at/all").should be_true
+      PrefixScope.prefix_covers?("/", "").should be_true
+    end
+
+    it "does not cover an unrelated url" do
+      PrefixScope.prefix_covers?("/api", "/admin/users").should be_false
+    end
+
+    it "handles prefixes with a trailing slash" do
+      PrefixScope.prefix_covers?("/web/", "/web/").should be_true
+      PrefixScope.prefix_covers?("/web/", "/web/x").should be_false
+    end
+
+    it "handles an empty url" do
+      PrefixScope.prefix_covers?("/web", "").should be_false
+    end
+
+    it "is case sensitive" do
+      PrefixScope.prefix_covers?("/web", "/WEB/x").should be_false
+      PrefixScope.prefix_covers?("/WEB", "/web").should be_false
+    end
+  end
+end
--- a/spec/unit_test/tagger/framework_taggers/prefix_scope_spec.cr
+++ b/spec/unit_test/tagger/framework_taggers/prefix_scope_spec.cr
@@ -19,15 +19,24 @@ describe PrefixScope do
     it "treats the root prefix as covering everything" do
       PrefixScope.prefix_covers?("/", "/anything/at/all").should be_true
       PrefixScope.prefix_covers?("/", "").should be_true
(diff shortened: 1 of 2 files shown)


hahwul:
Reviewed — the spec is exactly the right shape for a pure `(String, String) -> Bool`: no fixtures, no tagger state, and every assertion matches the implementation.

One thing worth changing rather than pinning: `prefix_covers?("/web/", "/web/x") == false` documents a real bug instead of intended behavior. Ktor (`normalize_prefix`) and Go (`join_prefix`) normalize before they reach here, but Express passes `match[1]` from `app.use('/admin/', requireAuth)` verbatim — and a trailing slash there is not a path boundary of its own, so that mount silently guarded nothing and every endpoint under it got reported as unguarded.

I pushed a commit that strips trailing slashes before the compare and flips those assertions to `be_true` (plus a url-side trailing slash and an empty-prefix case). Retitled the PR to `fix(tagger):` so the behavior change shows up in the release notes; your spec is what made it visible. Thanks!

PR #2668, opened 2026-08-28
refactor: replace plain `if !x.nil?` at remaining sites

### LINKED ISSUE #2651: Use `unless x.nil?` instead of `if !x.nil?` at the six remaining plain sites
An earlier sweep (#2169) replaced `if !x.nil?` with the idiomatic `unless x.nil?`
across the analyzers. Six plain single-condition sites are still around:

```console
$ grep -rnE '^\s*if !\w+\.nil\?$' src --include="*.cr"
src/miniparsers/import_graph.cr:622:        if !alias_name.nil?
src/analyzer/analyzers/python/sanic.cr:903:        if !support_methods.nil?
src/analyzer/analyzers/python/django.cr:1291:            if !method_function_match.nil?
src/analyzer/analyzers/python/django.cr:1768:        if !support_methods.nil?
src/analyzer/analyzers/python/flask.cr:1651:        if !support_methods.nil?
src/analyzer/analyzers/python/quart.cr:811:        if !support_methods.nil?
```

Each is a whole condition on its own, so `unless x.nil?` is a drop-in
replacement.

## What to do

Rewrite those six:

```crystal
# before
if !support_methods.nil?
# after
unless support_methods.nil?
```

The `else` branch, if any, keeps its meaning — `unless c ... else ... end` runs
the `else` when `c` is true. Check each site: if one has an `else`, prefer
`if x.nil? ... else ... end` over an `unless/else`, which the style guide
discourages.

### Out of scope

Do not touch the compound conditions. They are a different rewrite (`if match &&
match.size == 2`, relying on Crystal's nil narrowing) and belong in their own PR:

```console
src/analyzer/analyzers/python/flask.cr:426, 1417, 1587, 1593
src/analyzer/analyzers/python/quart.cr:722, 756, 760
src/analyzer/analyzers/python/fastapi.cr:1255
```

Also out of scope: `elsif !second.nil?` in `src/analyzer/analyzers/lua/lapis.cr`
(197, 231) — there is no `elsunless`, so those stay as they are.

## Verify

```bash
just check
crystal spec spec/unit_test
```

Nothing about endpoint output changes; `unless c` is `if !c`.

chei-l:
Closes #2651

### Description
Replaces plain `if !x.nil?` single-condition occurrences across the analyzers and miniparsers:
- At the 5 sites with an `else` branch (`import_graph.cr:622`, `sanic.cr:903`, `django.cr:1768`, `flask.cr:1651`, `quart.cr:811`), branches were inverted to `if x.nil? ... else ... end` to avoid `unless/else` per style guidelines (and Ameba's `Style/UnlessElse` rule).
- At `django.cr:1291`, the conditional is part of an `if ... elsif` chain. Because Crystal syntax does not support `unless ... elsif`, this was rewritten to `if method_function_match` to keep the control flow linear and avoid nested conditions, consistent with `any_method_function_match` in the subsequent `elsif`.

Compound conditions and `elsif !second.nil?` were left untouched as specified in #2651.

### Verification
- `just check` (format & Ameba lint) passed: 0 failures
- `crystal spec spec/unit_test` passed: 5571 examples, 0 failures
- `crystal spec spec/functional_test` passed: 24429 examples, 0 failures

### DIFF
--- a/src/analyzer/analyzers/python/django.cr
+++ b/src/analyzer/analyzers/python/django.cr
@@ -1288,7 +1288,7 @@ module Analyzer::Python
               current_http_methods = nil
             end
 
-            if !method_function_match.nil?
+            if method_function_match
               method_name = method_function_match[1].upcase
               current_http_methods = [method_name]
               suspicious_http_methods << method_name
@@ -1765,14 +1765,14 @@ module Analyzer::Python
       params.each do |param|
         is_supported_param = false
         support_methods = REQUEST_PARAM_TYPE_MAP.fetch(param.param_type, nil)
-        if !support_methods.nil?
+        if support_methods.nil?
+          is_supported_param = true
+        else
           support_methods.each do |support_method|
             if upper_method == support_method.upcase
               is_supported_param = true
             end
           end
-        else
-          is_supported_param = true
         end
 
         filtered_params.each do |filtered_param|
--- a/src/analyzer/analyzers/python/flask.cr
+++ b/src/analyzer/analyzers/python/flask.cr
@@ -1648,14 +1648,14 @@ module Analyzer::Python
       params.each do |param|
         is_support_param = false
         support_methods = REQUEST_PARAM_TYPES.fetch(param.param_type, nil)
-        if !support_methods.nil?
+        if support_methods.nil?
+          is_support_param = true
+        else
           support_methods.each do |support_method|
             if upper_method == support_method.upcase
               is_support_param = true
             end
           end
-        else
-          is_support_param = true
         end
 
         filtered_params.each do |filtered_param|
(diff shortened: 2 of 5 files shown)


hahwul:
Reviewed — no changes needed.

Checked each of the six sites against #2651: the five with an `else` were inverted to `if x.nil? ... else`, and the branch bodies are swapped correctly in all of them (`import_graph.cr` package_map, and the four `support_methods` filters in django/flask/quart/sanic — a nil lookup still means "supported"). Crystal narrows `support_methods` to non-nil in the `else`, so the `.each` needs no guard. `django.cr:1291` sits in an `if/elsif` chain where `unless` is not available, and plain `if method_function_match` is the right call there — it matches the `any_method_function_match` truthiness test in the next `elsif`, and `method_function_match[1]` still narrows.

Verified locally on the branch: format and Ameba clean, `crystal spec spec/unit_test` 0 failures, `crystal spec spec/functional_test` 24429 examples 0 failures. `grep -rnE '^\s*(if|elsif|unless) *! *\w+\.nil\?\s*$' src` now only returns the two `lapis.cr` `elsif` sites the issue put out of scope.

PR #2707, opened 2026-09-09
fix(python): compose Flask-AppBuilder routes from the class route base

hahwul:
## The defect

Flask-AppBuilder composes a route from a class-level base and a method-level suffix. noir read only the suffix:

```python
class ChartDataRestApi(ChartRestApi):
    resource_name = "chart"                 # superset/charts/api.py:208

    @expose("/data", methods=("POST",))     # superset/charts/data/api.py:254
    def data(self): ...
```

The URL is `/api/v1/chart/data`. noir reported `/data` — a false positive and a false negative in one endpoint. `superset/views/api.py` produced `/v1/time_range` for a route served at `/api/v1/time_range/`.

Found by hand-checking twenty of alibi's superset findings against the source: this cause accounted for six of the fourteen misses.

## The composition rule

Taken from the Flask-AppBuilder wheel Superset pins (5.2.2), cross-checked against 4.5.3 and 3.4.5 — unchanged in all three:

- `BaseView.create_blueprint` (`baseviews.py:281`): `route_base` defaults to `"/" + class_name.lower()`.
- `BaseApi.create_blueprint` (`api/__init__.py:562-566`): `resource_name` defaults to the lowercased class name, then `route_base` defaults to `f"/api/{version}/{resource_name}"`, with `version` defaulting to `"v1"`.

So `route_base` and `resource_name` are not the same knob: `route_base` replaces the whole prefix and never gets an `/api/<version>/`, while `resource_name` fills only the last segment. All three attributes are inherited across files.

The chain walk terminates on a 60-entry table of the Flask-AppBuilder classes that can serve as a base, carrying the attributes the library itself pins (`IndexView.route_base = ""`, `GroupApi.resource_name = "security/groups"`). **A class whose chain never reaches Flask-AppBuilder now emits nothing**, rather than a bare suffix.

Two collection defects surfaced in the same scan and are fixed: a multi-line `class Foo(\n    Mixin, Base\n):` header ended the class body at its own closing paren (losing all of `DashboardRestApi`), and a class written inside a docstring example was read as real code.

## Measurement

```
noir -b ~/oss-corpus/superset --only-techs python_flask
before 281 endpoints        after 318 endpoints
```

The number that matters is corroboration against Superset's own contract. Comparing the `python_flask` view with the `oas3` view of `docs/static/resources/openapi.json` (301 paths):

| | before | after |
| --- | ---: | ---: |
| corroborated | 140 | **204 of 301** |

Superset's full-scan total falls 550 → 523, because 64 more `python_flask` endpoints now coincide with `oas3` endpoints and merge. That drop *is* the corroboration.

**Controls:** netbox `python_django` 1210 and authentik `python_django` 3 with identical endpoint sets; every other Superset technology unchanged; the remaining eleven corpus repositories byte-identical.

**Fixture A/B**, 717 directories scanned one at a time, compared as full `(method, url)` sets rather than counts (5584 rows): one intended URL correction (`POST /api/v1/database` → `POST /api/v1/database/`) and four rows belonging to the fixtures this PR adds. Every other directory identical.

`just test` 5727 + 15244 examples, 0 failures.

## Deliberately not done

**Synthesising Flask-AppBuilder's generated CRUD and `_info` routes.** Measured the counterfactual: emitting `GET/POST base/`, `GET/PUT/DELETE base/{pk}` and `GET base/_info` for every `ModelRestApi` subclass in Superset produces 108 pairs, of which **68 are in the OpenAPI document and 40 are not** — a 37% invented-URL rate. The reason is `include_route_methods` / `exclude_route_methods`, whose values in real projects are computed set expressions over project-local constants (`ChartRestApi.include_route_methods = RouteMethod.REST_MODEL_VIEW_CRUD_SET | {…}`, where `RouteMethod` is Superset's own class). noir cannot evaluate that, and five Superset APIs restrict it so the CRUD set does not exist on them. Emitting URLs that 404 is the exact defect this PR fixes.

The same argument rules out **propagating a base class's `@expose` handlers to subclasses**: `ChartDataRestApi.include_route_methods = {"get_data", "data"}` means none of `ChartRestApi`'s 22 routes are registered on it, so blind propagation would invent ~40 phantom `/api/v1/chart/…` duplicates.

**A registration filter** (`appbuilder.add_api(X)` / `add_view(X)`) would remove the one remaining false positive precisely, but only 61 of Superset's 83 `@expose`-carrying classes appear in a literal `add_*(ClassName)` call — the other 22, including `SupersetIndexView` and `DatabaseView`, are registered through the security manager or through lists. It would cost 20+ true endpoints to remove 2 false ones.

**Deriving the kind from a `*Api` / `*View` class-name suffix** when the chain does not reach Flask-AppBuilder: still a guess, and after the built-in table landed every one of Superset's 83 classes resolves without it.


hahwul:

## What is blocking it

```
```

PR #2708, opened 2026-09-09
feat(javascript): read Express routes registered through a forwarding helper

hahwul:
## The defect

NodeBB registers its entire HTTP surface through three helpers in `src/routes/helpers.js` — `setupApiRoute`, `setupPageRoute`, `setupAdminPageRoute` — rather than calling `router.get(...)` at the call site, and mounts the write API from a module whose router arrives as a function parameter. noir saw neither, so **218 documented `/api/v3/*` paths had no code counterpart at all**.

Two separate defects, fixed in two commits, because fixing only one produces routes at the wrong prefix:

**The mount.** `RouterMountScanner` refuses to place `router.use('/api/v3/users', …)` until it knows where `router` itself is mounted. In `src/routes/write/index.js` the router is destructured from a function parameter (`Write.reload = async (params) => { const { router } = params; … }`), so the mount was deferred forever and dropped. A third pass now places the leftovers at the root — but only onto children no other mount reached.

**The helper.** A new `RouteHelperScanner` indexes functions whose body forwards to `<recv>[<verb>](<path>, …)` with the receiver and path bound to their own parameters (directly, or through a positionally destructured rest parameter), then reads the literal verb and path off every call site, following cross-file helper imports and taking the prefix from the same `CodeLocator` table the generic route pass uses.

## Measurement

```
noir -b ~/oss-corpus/NodeBB --only-techs js_express
before 115 endpoints, 0 under /api/v3
after  567 endpoints, 220 under /api/v3
```

All 453 helper-derived additions trace to a `setup*Route(` call at the reported file:line; all 218 write-API routes trace their prefix to the exact `router.use` that mounts their module.

The measurement that matters is how much of NodeBB's own contract now has an implementation. Of its 368 documented (`oas3`) paths:

| | before | after |
| --- | ---: | ---: |
| strict `(method, url)` match | 14 | **247** |
| segment-wildcard match | 70 | **366 of 368** |

**Controls:** the other eleven corpus repositories and all 725 pre-existing fixture directories are byte-identical.

## False positives were the whole design constraint

A prior audit of this same repository found ~250 "endpoints" that were HTTP *client* calls in test files. So acceptance is deliberately narrow, and it was tested against a hostile synthetic project whose `src/http-client.js` holds five forwarding functions — including one structurally identical to `setupApiRoute`:

```js
exports.request = function (client, method, url) { return client[method](url, { timeout: 5000 }); };
```

**The first version of the scanner minted three routes from it.** That project is now committed as a negative fixture, and it yields only its one real route.

Seven independent acceptance conditions, each with the case it exists to kill:

| condition | kills |
| --- | --- |
| receiver, path and verb are parameters of the same function | `request.get(url)` inside `function fetch(url)` |
| receiver parameter is router-named (`router`, `app`, `server`, `*Router`…; `api` deliberately excluded) | `client[method](url, opts)`, `api[method](url, body)` on an axios instance |
| forwarded call has ≥1 argument after the path | one-argument `client.get(url)` |
| ≥1 forward passes the path parameter bare | helpers the generic parser already handles — indexing them both duplicated and suppressed those routes |
| verb and path are literals at the call site | `setupApiRoute(router, 'get', buildPath('exports'), …)` |
| comments blanked first, offset-preserving | a commented-out `// setupApiRoute(router, 'post', '/', …)` at `src/routes/write/search.js:14` |
| test trees, minified files and test stubs skipped | a helper call from `test/api.js` |

The last two were caught by hand-verifying a sample rather than by the corpus; both are fixtures now.

## Cost

One extra regex scan per JS/TS file, plus comment-stripping and a call-site scan only on files that can actually reach an indexed helper: +5% on directus, +13% on NodeBB — where the output is 4.9× larger.

## Known gaps, not fixed here

`${name}` is not propagated into NodeBB's page routes: `src/routes/index.js` calls `_mounts.admin(router, mounts.admin, …)` where `mounts.admin` is `'admin'`, and `src/routes/admin.js` registers `` `/${name}/dashboard` ``. noir emits `/api/{name}/dashboard` where the doc says `/api/admin/dashboard` — 121 of the 368 documented paths fail strict matching for only this reason. Propagating a literal argument into a callee's template literal is `js_parser.cr` territory and should be designed together with this helper index rather than raced against it.

`RouteHelperScanner` reads only file-level mount prefixes, not the function-scoped ones `RouterMountScanner` also writes; no corpus project hits the difference. And the index is Express-only — Hono, Koa and Fastify run the same shared extractor and would benefit, but no corpus project uses a forwarding helper with them, so there is nothing to measure yet.


hahwul:

## What is blocking it

```
```

PR #2709, opened 2026-09-09
perf: search substrings with memchr and pre-gate detector regexes on a required literal

hahwul:
Two changes, both detection-neutral by construction.

```
twelve-repo corpus   12.26 s -> 8.27 s   (1.48x)
superset              3.60 s -> 1.70 s   (2.12x)
```

Independently re-measured against a release build of current main, median of five runs each: superset 2.32×, netbox 2.85×, authentik 2.04×, directus 1.36×, kong 1.25×, casdoor 1.17×, gitea 1.12×.

## `memchr` instead of Rabin-Karp

Crystal's `String#index(String)` uses Rabin-Karp, which reads every byte of the haystack and maintains a rolling hash. noir's hot path is thousands of `content.includes?("...")` gates over whole source trees, where the needle is almost always absent.

`src/ext/string_search.cr` picks the **rarest byte of the needle** (by a static English/source-code byte-frequency table) and `memchr`s for that, only then comparing the full needle. A miss now costs one SIMD scan per candidate byte rather than a rolling hash over the file. Every `includes?` gate in the codebase gets faster without any of them changing.

## A literal pre-gate for 31 detectors

An import-marker regex like `/^\s*import\s+(?:foo|bar|baz)/m` has to be run over the whole file to fail. But every alternative in these detectors shares a literal that must appear for any of them to match. Testing that literal first with the substring search above turns a failing regex pass over a whole source tree into a single `memchr` sweep.

Superset's detect phase falls from **2.30 s to 0.52 s**.

## Detection is unchanged

- All 725 fixture directories, each scanned separately: byte-identical.
- All twelve corpus repositories: normalized JSON byte-identical (verified independently by the reviewer as well — 12/12).
- `crystal spec` 5764 + 15239 examples, 0 failures. `just check` (format + ameba) clean.

## Built, measured, and reverted

Four optimisations did not survive their own measurement:

1. **Memoising `JSRouteExtractor.other_shared_extractor_framework?`** — profiling put it at the top of the Express analyzer, called twice per file with identical arguments. The memo was detection-neutral but **not faster**: superset +4.2%, NodeBB +5.6%, kong +5.2%. The cache key hashes the whole file, and that costs about what the saved regex passes cost.
2. **A hand-rolled Flask comment-strip** replacing the `gsub` — measured first: over all 2661 superset Python files (25.9 MB) the regex costs 18.7 ms and a byte-exact hand-rolled equivalent 4.96 ms. Saving 14 ms on a 1.7 s scan does not justify new hand-written string code.
3. **Interning `TreeSitter.node_type`** — exclusive-time profiling put `ts_node_type` + `String.new` + `strlen` at ~3.5% of real work, which does not justify the thread-safety machinery an interning table needs against concurrent analyzer fibers.
4. **Caching tree-sitter trees in `TreeSitter.parse`** — this would be the real fix for the remaining minio outlier, but `extraction_result_cache.cr` documents the opposite decision (a tree's lifetime is tied to its parse block), and doing it safely needs refcounting so one fiber cannot free a tree another is walking. Too large to land under a detection-neutral-or-nothing bar.

## Remaining outlier

minio barely moves (1.03×). Its cost traces to redundant tree-sitter parses in the Go analyzers — the same files parsed more than once across passes. That is item 4 above and is handed back rather than attempted here.


hahwul:

## What is blocking it

```
```

PR #2710, opened 2026-09-10
ci: stop the spec binary exporting every Crystal symbol

hahwul:
`bin/noir_spec` was linking successfully and then dying the moment it was executed:

```
```

## The message is not a corrupt binary

Crystal passes `-rdynamic` so backtraces can symbolize, which exports **every** Crystal symbol into `.dynsym`. For this binary, `readelf` reports:

```
Version symbols section '.gnu.version' contains 133142 entries
STRTAB (.dynstr)                4,921,242 bytes
```

`.gnu.version` entries are 16-bit indices into the version table. At this size the loader read an out-of-range index and dereferenced into `.dynstr` — which is 4.9 MB of Crystal mangled names. So a libc or libxml2 symbol came out carrying a *Crystal method signature* as its ELF symbol version, and the pair differed per build (`stderr` with `Hash(K, V)#entries_full?:Bool` on one branch, `xmlFree` with `Hash(K, V)#fetch` on another and on main).

Two more readings from the same diagnostic run confirm the scale: `readelf -V` took **13 minutes** on this binary, and `readelf --dyn-syms | grep -E 'xmlFree|stderr'` took another 13 and matched **nothing** — the symbols the loader named are not in the table it was reading.

`ldd` resolves every library including `libxml2.so.2` (2.9.14), so nothing was missing from the image.

## The change

`ld` takes the last of `--export-dynamic` / `--no-export-dynamic`, so appending the latter through `--link-flags` cancels the one Crystal adds. Nothing dlopens this binary.

A guard step fails the job if the count grows back above 2,000, because the symptom points nowhere near the cause and cost several rounds to find.

## Why not split the binary instead

Halving the symbol count by building unit and functional separately would work, but #2699 established that compiling `src/` is essentially the entire cost and does not grow when the suites are combined — unit alone and unit-plus-functional both compile in 20.9 s. Splitting would pay that twice and undo a measured optimization to work around a linker flag.

## Eliminated on the way here, each by measurement

1. **Not any one pull request** — it reproduced on `main`: one of four runs of the `fix(lua_lapis)` merge commit failed this way.
3. **Not the tree-sitter object cache** — after deleting the entry and keying on the runner image, the run rebuilt every grammar from source and reported `no previous .o files were reused`. A fully cold build failed the same way.
4. **Not `--no-debug`** — dropping it changed nothing, which is expected: debug info does not live in `.dynsym`.
5. **Not a race between the four concurrent invocations** — all four failed identically.
6. **Not the source** — `just test` on a merge of current main passes on each blocked branch: ~5,730 unit + ~15,240 functional examples, 0 failures.

Those first three attempts are reverted on this branch; the diff against main is the build flag and the guard.

## Unblocks

#2707, #2708 and #2709, each green on every check except this job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,13 +175,10 @@ jobs:
           # Crystal fingerprints its own cache entries by content, so a
           # partially-stale restore is safe — it only ever means some entries
           # get recompiled. Hence the loose restore-keys ladder.
-          # The ladder stops at the `-build-` prefix: the bare
-          # `crystal-<os>-<version>-` prefix reaches the spec and lint jobs'
-          # caches, which are compilations of different programs. See the spec
-          # job's cache step for what that produced.
-          key: crystal-${{ runner.os }}-${{ matrix.crystal-version }}-build-v2-${{ hashFiles('shard.lock', 'src/**/*.cr') }}
+          key: crystal-${{ runner.os }}-${{ matrix.crystal-version }}-build-${{ hashFiles('shard.lock', 'src/**/*.cr') }}
           restore-keys: |
-            crystal-${{ runner.os }}-${{ matrix.crystal-version }}-build-v2-
+            crystal-${{ runner.os }}-${{ matrix.crystal-version }}-build-
+            crystal-${{ runner.os }}-${{ matrix.crystal-version }}-
       - name: Install dependencies
         run: shards install
       - name: Build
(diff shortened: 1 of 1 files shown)


hahwul:
Closing. Neither of the two changes I attempted here fixed the `symbol lookup error` — the Crystal cache scoping and the tree-sitter image keying were both disproven by measurement, and the second is already reverted on this branch. What is left is unvalidated CI hygiene, and adding that to a CI system that is currently misbehaving in a way nobody understands is risk without benefit.

The diagnosis stays readable in the description above, and the same summary is on #2707, #2708 and #2709, which are blocked on the identical job. Branch kept.

PR #2744, opened 2026-09-17
fix(csharp): correct ASP.NET Core MVC controller discovery

### LINKED ISSUE #2743: ASP.NET Core MVC: controller discovery misses valid actions and includes commented-out routes
## Summary

The ASP.NET Core MVC analyzer misses valid controllers/actions and reports commented-out actions. Reproduced on current `main` (`6b69d09fbe62bf25eb2187c01469f480da986c08`) with newly authored, source-only fixtures; no application build or execution is needed.

### 1. Discovery assumes a Controller filename/class suffix

Put this in `Endpoints.cs` in an ASP.NET Core project:

```csharp
using Microsoft.AspNetCore.Mvc;

[ApiController]
[Route("catalog")]
public class Catalog : ControllerBase
{
    [HttpGet]
    public IActionResult List() => Ok();
}
```

Expected: `GET /catalog`. Actual: no endpoint. The analyzer filters filenames containing `Controller` and then matches only class names ending in `Controller`. ASP.NET Core also discovers types with a controller attribute, including an inherited attribute; `ControllerBase` supplies that marker. A suffix-named controller in an arbitrarily named source file is missed too.

Discovery should still exclude `[NonController]`, non-public, abstract, open-generic, and nested types. Inheriting a name suffix alone does not confer a controller attribute.

### 2. NonAction leaks from a private helper to the next public action

Inside a routed controller:

```csharp
[NonAction]
private IActionResult Helper() => Ok();

[HttpGet("live")]
public IActionResult Read() => Ok();
```

Expected: the public action is emitted. Actual: it is suppressed. The pending `NonAction` state is only consumed when the analyzer reaches a public signature, so a private/protected helper leaves it behind.

### 3. Commented-out actions become endpoints

Inside a routed controller:

```csharp
// [HttpDelete("removed")]
// public IActionResult Removed() => Ok();
```

Expected: no endpoint. Actual: a DELETE endpoint is emitted. Block comments and XML documentation examples can likewise be interpreted as actions. The extraction loop reads raw lines even though `CSharpLexer` already exposes comment-stripped source with preserved locations.

## Reproduction/validation

A functional regression fixture covering these cases plus discovery exclusions currently produces **14 examples, 12 failures, 0 errors** with Crystal 1.21.0. I am preparing a focused PR with the synthetic fixtures and fixes.

Reference semantics: https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/Controllers/ControllerFeatureProvider.cs

peterkarman1:
## Summary

Fixes #2743.

Correct three ASP.NET Core MVC extraction defects without building or executing the scanned application:

- Discover controllers from source types rather than filenames. Recognize the `Controller` suffix and explicit/inherited controller markers, including locally resolvable base types and the framework's `ControllerBase`/`Controller` bases. Keep inheritance lookup project-scoped and reject ambiguous local base names. Exclude non-public, abstract, open-generic, nested, and `[NonController]` types; a base type's name suffix alone does not make its subclasses controllers.
- Consume pending action attributes at private/protected as well as public method boundaries, so `[NonAction]` on a helper no longer suppresses the following public action.
- Use the existing C# lexer's comment-stripped source and structural brace information. Commented attributes/actions are not emitted, actions stay within their declaring class, and original endpoint/callee line numbers are preserved.

A small syntax-only type extractor lives in `src/miniparsers/`; MVC discovery semantics remain in the framework adapter. This is source-level discovery, not assembly loading or complete C# symbol resolution.

## Regression coverage

All fixtures are newly authored synthetic examples. They cover arbitrarily named files, multiple classes per file, explicit and inherited controller markers, discovery exclusions, private/protected/public `[NonAction]`, and line/block/XML-documentation comments.

Before the fix on `6b69d09f`: **14 examples, 12 failures, 0 errors**.
After the fix: **14 examples, 0 failures, 0 errors**.

## Validation

Using Crystal 1.21.0 in the documented Alpine container:

- `shards build` — passed.
- `crystal spec spec/functional_test/testers/csharp/*_spec.cr` — **565 examples, 0 failures, 0 errors**.
- Crystal format check and Ameba on all changed Crystal files — passed.
- Full `spec/suite.cr` run — **21,025 examples, 1 failure, 1 error, 8 pending**. Both failures reproduce on untouched upstream `6b69d09f` in the same container:
  - `config_initializer_spec.cr:415`: running as root can still read a chmod-000 file.
  - `detector/silent_loss_reporting_spec.cr:261`: Alpine's current-directory lookup raises `Filename too long` while constructing the beyond-PATH_MAX fixture.

No changes to those unrelated tests, dependencies, or other framework analyzers.


hahwul (approved):
LGTM

hahwul:
HI @peterkarman1 
Great work! Thanks for the contribution :)

